PR #14, opened 2019-12-19
Chore: Minor enhancements

facuspagnuolo:


package.json
```diff
@@ -1,6 +1,8 @@
 {
```
sohkai: As a note on this, we should add a LICENSE (we can copy https://github.com/aragonone/voting-connectors/blob/master/LICENSE)

PR #30, opened 2020-01-09
Server: add production settings

sohkai:


server.Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM node:12.14.0-alpine
```
bingen: As this is only for the server, I think the Dockerfile should be in that folder.
Besides, we should add this image to docker-compose file there to orchestrate it with PostgreSql image for db.

sohkai: For ease of k8s deployments, we decided to leave it here.

facuspagnuolo:
I think we already merged all this content in different commits, let me know if we can close it @sohkai

PR #31, opened 2020-01-10
App: deploy to now

izqui:


sohkai:
Realizing now that for `now` to build this app, we'll have to publish `@aragon/court-backend-shared`, which is a bit annoying. Will resume this later.

sohkai:
The frontend is now being deployed directly onto k8s along with the rest of the backend and services.

PR #46, opened 2020-01-23
Minor fixes

bingen:


packages/services/bin/worker.js
```diff
@@ -5,21 +5,22 @@ import errorHandler from '../src/helpers/error-handler'
...
 
 Logger.setDefaults(false, true)
 
-const logger = Logger(name)
+const logger = Logger(name, prefixColor)
 const worker = require(path.resolve(process.cwd(), workerPath)).default
 
 async function run() {
-  logger.info(`Starting worker, will run ${times} jobs`)
+  logger.info(`Starting worker, will run ${times == 0 ? 'infinite' : times} jobs`)
```
sohkai: This sounds like a bug... should we change the next line to `let job = 0` instead?

bingen: Why? I don't get it. I think the idea is if `times == 0`, then the loop will never exit, as `job` cannot be zero, and therefore `times = 0` means infinite. What am I missing?

PR #48, opened 2020-01-27
app: Add heartbeat button

bingen:


packages/app/src/components/court/CourtConfig.react.js
```diff
@@ -62,6 +65,7 @@ export default class CourtConfig extends React.Component {
         <p>Term duration: {bn(termDuration).div(bn(60)).toString()} minutes</p>
         <p>Current term: #{currentTerm}</p>
         <p>Needed transitions: {neededTransitions.toString()}</p>
+        <p><span onClick={() => this._heartbeat()} style={{cursor: 'pointer', textDecorationLine: 'underline'}}>Heartbeat</span></p>
```
sohkai: We could probably use a `<a>` here and then we could avoid the custom styling (no `href` will disable linking).

bingen: Sure, I was just following the style on `claim` buttons, but this makes much more sense.

PR #51, opened 2020-01-29
server/services: Remove errors from DB

bingen:
And log to stdout instead.


packages/server/src/database/migrations/20200102112832-create-reveals.js
```diff
@@ -42,15 +42,6 @@ module.exports = {
...
         },
-        errorId: {
-          allowNull: true,
-          type: Sequelize.INTEGER,
-          references: {
-            model: 'ErrorLogs',
-            key: 'id',
-            as: 'errorId',
-          },
-        },
```
facuspagnuolo: For future changes, this is not how migrations are supposed to work, otherwise we will need to update already-created databases manually

bingen: What do you mean? We discussed with @sohkai and decided to remove errors from database.

facuspagnuolo: Not saying about the change per se but the way it was implemented. Migrations are supposed to work as changes for a database, if we change a migration file in order to perform a change on the DB it won't affect the DBs that already ran that migration, you're supposed to create a new migration for that change instead

bingen: Oh, now I get it, thanks. Yes, ok, but this is not "on production" yet, so we'll create the db from scratch.

PR #53, opened 2020-02-04
cli: Add config command

bingen:


packages/cli/src/commands/config.js
```diff
@@ -0,0 +1,60 @@
+const path = require('path')
+const logger = require('@aragon/court-backend-shared/helpers/logger')('config')
+
+const command = 'config'
+const describe = 'Call court hearbeat'
```
sohkai: Should this be `Change court config`?

PR #55, opened 2020-02-07
Dockerize server and services

bingen:
Not sure how this with play out with k8s, but locally it works just by doing: `docker-compose down && docker-compose build && docker-compose up`


packages/server/docker-entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+# TODO: wait for db server
+sleep 30
+
+npx sequelize db:create
```
promaty: This guy seems to prevent container restarts, because it throws an error if database already exists

PR #57, opened 2020-02-11
Services: Remove unused stuff

facuspagnuolo:


packages/services/src/workers/reveal.js
```diff
@@ -1,4 +1,3 @@
-import { Op } from 'sequelize'
```
bingen: :+1: This had been removed [here](https://github.com/aragonone/court-backend/pull/55), but yes, better in an independent PR.

PR #62, opened 2020-02-11
App: Polish heartbeat button and extend jurors list

facuspagnuolo:


packages/app/src/actions/court.js
```diff
@@ -2,6 +2,8 @@ import Network from '../web3/Network'
 import ErrorActions from './errors'
 import * as ActionTypes from '../actions/types'
 
+const HEARTBEAT_MAX_TRANSITIONS = 10
```
sohkai: cc @bingen I remember we had a conversation about increasing this to `20`, since it would very safely fit under the gas limit.

packages/app/src/actions/court.js
```diff
@@ -78,11 +80,11 @@ const CourtActions = {
     }
   },
 
-  heartbeat(transitions) {
+  heartbeat() {
```
sohkai: Wonder if it would still be useful to make this configurable; perhaps the CLI would have a flag to say how many transitions it would allow.

facuspagnuolo: Note that it is (see https://github.com/aragonone/court-backend/pull/62/files#diff-5545618aa375119dd0b2348de4adc9a7R87), I think you're mixing the frontend action which is not configurable for the moment with the backend helper

sohkai: Ah right, gotcha. Nevermind then!

packages/app/src/actions/court.js
bingen: Yes, I think it was in the context of [services](https://github.com/aragonone/court-backend/blob/master/packages/services/src/workers/heartbeat.js#L5)

PR #68, opened 2020-02-12
Shared: Unify GraphQL client usage

facuspagnuolo:


packages/server/.env.sample
```diff
@@ -1,16 +1,21 @@
...
+
+# Admin config
 SECRET_KEY=secret
-CORS_WHITELIST=https://anj.aragon.org
-ADMIN_EMAIL=facu@aragon.one
+ADMIN_EMAIL=you@aragon.one
 ADMIN_PASSWORD=aragon
 
-# Web3 provider
-#PRIVATE_KEY=
```
bingen: I did this on purpose, to raise awareness that you need this variable, although I was passing it through command line.

PR #73, opened 2020-02-13
Server: Normalize users subscriptions

### LINKED ISSUE #103: Server: Implement endpoint to check if a subscription exists

facuspagnuolo:
Fixes #103 

I recommend reviewing by commit

Some things we should discuss/do after this is merged:
- We should update the microsite or any other website subscribing users to check first and create if they weren't registered
- We should probably want to do some security check like requesting the user to sign a message to ensure they hold the address they are trying to register
- We may want to add another security check like an email verification step


packages/server/src/validators/UsersValidator.js
```diff
@@ -0,0 +1,27 @@
...
+
+class UsersValidator extends BaseValidator {
+  async validateForCreate({ email, address }) {
+    this._validateEmail(email)
+    await this._validateAddress(address)
+    return this.resetErrors()
+  }
+
+  _validateEmail(email) {
+    if (!email) this.addError({ email: 'An email address must be given' })
```
sohkai: Let's do a bit more of a sanity check here for their email :).

[`joi`](https://www.npmjs.com/package/joi) is quite nice, although it's more intertwined with `hapi`. I'm sure we can find something equivalent for other frameworks that should allow us to declare types just as easily.

packages/server/src/validators/UsersValidator.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  async _validateAddress(address) {
+    if (!isAddress(address)) return this.addError({ address: 'Given address is not valid' })
```
sohkai: Does `isAddress()` check for the checksum?

packages/server/src/validators/UsersValidator.js
```diff
@@ -0,0 +1,27 @@
...
+
+    const checksumAddress = toChecksumAddress(address)
```
sohkai: What you think about storing addresses in the db as just lowercased addresses?

I've been thinking about this for a bit and more or less decided that it would be better if we:

- Sanity checked all user-generated addresses with checksums
- Only used lowercase addresses internally, once those checks have passed

Forgetting to do the `addressEquals()` check instead of a `===` check has bit people way too many times, and it's because we don't have strong responsibility rules around when we preserve and when we strip the checksum.

packages/server/src/controllers/UsersController.js
```diff
@@ -0,0 +1,49 @@
...
+import UsersValidator from '../validators/UsersValidator'
+import { toChecksumAddress } from 'web3-utils'
+
+const { User, UserAddress } = Models
+
+export default {
+  async exists(request, response, next) {
+    try {
+      const errors = await UsersValidator.validateForCreate(request.query)
+      return (errors.length === 0) ? response.status(200) : response.status(400).send({ errors })
```
sohkai: I might be missing context, but it seems kind of weird we have a pattern where we short circuit with a `400` inside, but if the method throws, we pass the error along to `next()`.

packages/server/src/controllers/UsersController.js
```diff
@@ -0,0 +1,49 @@
...
+      response.status(200).send(user)
+    } catch(error) {
+      next(error)
+    }
+  },
+
+  async all(request, response, next) {
+    try {
+      const limit = request.query.limit || 20
+      const offset = (request.query.page || 0) * limit
```
sohkai: Is there a way for an API consumer to know how many pages there are?

packages/server/src/database/migrations/20200212131402-normalize-users.js
```diff
@@ -0,0 +1,127 @@
...
+        } else {
+          const { id: addressId, userId: addressUserId } = existingAddress[0]
+          logger.warn(`Skipping address '${parsedAddress}' for user with ID '${userId}', another user with ID '${addressUserId}' already has that address with ID '${addressId}'`)
+        }
+      }
+    }
+
+    console.log('')
+    const usersWithoutAddress = (await queryInterface.sequelize.query(`SELECT email FROM "Users" LEFT JOIN "UserAddresses" ON "Users"."id" = "UserAddresses"."userId" WHERE "UserAddresses"."id" IS NULL`))[0]
+    logger.warn('WARNING! The following email do not have an email associated:')
```
sohkai:
```suggestion
    logger.warn('WARNING! The following emails do not have an address associated:')
```

But this should never be the case, right?

packages/server/src/database/migrations/20200212131402-normalize-users.js
```diff
@@ -0,0 +1,127 @@
...
+      else {
+        const parsedAddress = isAddress(address) ? toChecksumAddress(address) : address
+        const existingAddress = (await queryInterface.sequelize.query(`SELECT * FROM "UserAddresses" WHERE address='${parsedAddress}'`))[0]
+
+        if (existingAddress.length === 0) {
+          logger.success(`Inserting new address '${parsedAddress}' for user with ID '${userId}' with email '${parsedEmail}'`)
+          await queryInterface.sequelize.query(`INSERT INTO "UserAddresses" ("userId", "address", "createdAt", "updatedAt") VALUES ('${userId}', '${parsedAddress}', '${createdAt.toUTCString()}', '${updatedAt.toUTCString()}')`)
+        } else {
+          const { id: addressId, userId: addressUserId } = existingAddress[0]
+          logger.warn(`Skipping address '${parsedAddress}' for user with ID '${userId}', another user with ID '${addressUserId}' already has that address with ID '${addressId}'`)
```
sohkai: I think there might be another case here where it's because the address has already been added to the `UserAddresses` table—should we check against `existingAddress.some(address => address === parsedAddress)`?

packages/server/src/database/migrations/20200212131402-normalize-users.js
```diff
@@ -0,0 +1,127 @@
...
+      }
+      else {
+        const { id: userId } = existingUser[0]
+        logger.warn(`Skipping email '${parsedEmail}', user with ID '${userId}' already has that email address`)
+      }
+
+      const { id: userId } = (await queryInterface.sequelize.query(`SELECT id FROM "Users" WHERE email='${parsedEmail}'`))[0][0]
+      if (!address) logger.warn(`Skipping address '${address}' for user with ID '${userId}'`)
+      else {
+        const parsedAddress = isAddress(address) ? toChecksumAddress(address) : address
```
sohkai: I wonder here, should we skip and log and error if the address doesn't parse to a correct checksum address?

packages/server/src/validators/UsersValidator.js
facuspagnuolo: If the address provided is checksumed, it does

packages/server/src/validators/UsersValidator.js
facuspagnuolo: Sounds great, let's do that!

packages/server/README.md
```diff
@@ -41,31 +41,42 @@ Remember to append `NODE_ENV=production` if needed.
...
     ```
 
-#### 1.2. All
+#### 1.2. Exists
 
-- URL: /subscriptions
+- URL: /user
+- Method: GET
+- Query: 
+  - `email`: Email address, string mandatory
```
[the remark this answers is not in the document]
facuspagnuolo: Totally, thx for the suggestion

packages/server/src/database/migrations/20200212131402-normalize-users.js
facuspagnuolo: It shouldn't but logically possible :)

packages/server/src/validators/UsersValidator.js
sohkai: As discussed afterwards, it would most likely be best to always validate the checksum since this user generated information.

packages/server/src/database/migrations/20200212131402-normalize-users.js
facuspagnuolo: The problem is that if the address was already added, it means it has an email associated. I think we must support multiple addresses per email, but not multiple emails per address, i.e. 1-N email-address relation. That's exactly why [this case](https://github.com/aragonone/court-backend/pull/73#discussion_r398966540) is possible. If we find the same address linked to multiple emails, we will have the first email linked to the address, but the rest will be left without one.

packages/server/src/database/migrations/20200212131402-normalize-users.js
facuspagnuolo: Actually, it can happen, see my answer [here](https://github.com/aragonone/court-backend/pull/73#discussion_r399263599)

packages/server/src/database/migrations/20200212131402-normalize-users.js
facuspagnuolo: Definitely, was about to fix that :)

packages/server/src/controllers/UsersController.js
facuspagnuolo: We are returning the total number of users as well, you can get it dividing the limit you want to request, would it be better to return the number of pages instead?

packages/server/src/validators/UsersValidator.js
facuspagnuolo: Yea, we [are using sequelize validations](https://github.com/aragonone/court-backend/pull/73/files#diff-402aeeda5eda2223bd3da4bf15d0c2bbR10) for this, but let's pre-validate here as well

packages/server/src/controllers/UsersController.js
facuspagnuolo: I was trying to follow [this guide](https://expressjs.com/en/guide/error-handling.html), but maybe there is a better way to handle errors

packages/server/src/controllers/UsersController.js
sohkai: Hadn't noticed that—that sounds good!

packages/server/src/controllers/UsersController.js
sohkai: Yes, I'm not entirely familiar with how the middleware works when you use `response.send()` like this, but I assume it'll short circuit?

Anyhow, if we need to, we can refactor how we do this input validation in a different PR.

facuspagnuolo:
Yes! Merging it will deploy it to all the testing envs

PR #74, opened 2020-02-13
Chore: Update root readme

facuspagnuolo:


.env.sample
```diff
@@ -1,16 +1,21 @@
+# Server config
 PORT=8000
+CORS_WHITELIST=https://anj.aragon.org
```
sohkai: Could we add `court.aragon.org` to this list? I think it should just be a `,https://court.aragon.org` but configuring arrays via `.env`s always take a bit to figure out for me.

packages/server/.gitignore
```diff
@@ -1,2 +1 @@
-.env
```
bingen: Why are we removing this from here and `services` if still have those files?

facuspagnuolo: cause now we have a `.env` in the root dir and I moved the ignore statement there

PR #77, opened 2020-02-13
CLI: Support precedence campaign Arbitrable

facuspagnuolo:


packages/app/src/actions/accounts.js
```diff
@@ -47,10 +47,12 @@ const AccountActions = {
     return async function(dispatch) {
       try {
         const ant = await Network.getANT()
-        const symbol = await ant.symbol()
-        const antBalance = await ant.balanceOf(account)
-        const balance = fromWei(antBalance.toString()) // TODO: assuming 18 decimals
-        dispatch(AccountActions.receiveAntBalance({ symbol, balance, address: ant.address }))
+        if (ant) {
```
sohkai: Maybe we should error if we can't find the ANT contract

facuspagnuolo:
> It seems a bit weird to hardcode our precedence campaign Arbitrable into the CLI like this, but let's make it generic when we need to (perhaps with Agreements).

Yea, I agree, but given that we are the only ones using these commands for now, I think we are fine until we need these to work in a general way. As you said, that will probably happen when we start interacting with Agreement instances

PR #89, opened 2020-03-09
Chore: Improve dockerization

### LINKED ISSUE #35: Service: deploy to testnet

### LINKED ISSUE #36: Service: deploy to mainnet

facuspagnuolo:
Specially thanks to @promaty and @bingen :)

Fixes #35
Fixes #36


.github/workflows/mainnet.yml
```diff
@@ -37,6 +39,8 @@ jobs:
         kubectl config set-credentials aragon --token=$(base64 -d <<< ${{ secrets.KUBE_TOKEN }})
         kubectl config set-context aragon --cluster=aragon --user=aragon
         kubectl config use-context aragon
-    - run: kubectl set image deployment/court-backend server=$REPO:${GITHUB_REF/*server-/}
+    - run: kubectl set image deployment/court-backend server=$REPO:${GITHUB_REF}
+    - run: kubectl set image deployment/court-backend services=$REPO:${GITHUB_REF}
+    - run: kubectl set image deployment/court-backend-rinkeby app=$REPO:${GITHUB_REF}
```
bingen: Why do we have Rinkeby here?

Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM node:12.14.0-alpine
+RUN apk add --no-cache git
+
+WORKDIR /court-backend
+
+# copy root package and lerna json files
+ADD ./package.json /court-backend/package.json
```
bingen: Curious: any particular reason to use `ADD` instead of `COPY`?

docker-compose.yml
```diff
@@ -8,20 +8,41 @@ services:
...
     image: aragon/court-server
+    command: npm run start:server:dev
     depends_on:
       - court_server_db
     ports:
-      - "8080:8080"
-      - "8081:8081"
+      - "8000:8000"
+    volumes:
+      - ./packages/server/src:/court-backend/packages/server/src
```
bingen: Nice addition! I guess this is for "hot loading". I wonder if it would be useful to have `build` folder too.

docker-compose.yml
```diff
@@ -8,20 +8,41 @@ services:
...
+
+  court_backoffice:
+    build: .
+    image: aragon/court-backoffice
+    command: npm run start:app
+    depends_on:
+      - court_server
+    ports:
+      - "3000:3000"
+      - "35729:35729"
```
bingen: What's this one for?

packages/server/scripts/db-setup.sh
```diff
@@ -4,7 +4,6 @@
 db_result=0
 while [ "${db_result}" -ne "1" ]; do
     echo "Waiting for DB..."
-    # npx sequelize db:migrate:status 2>&1 | grep 'connect ECONNREFUSED' > /dev/null
     npx sequelize db:migrate:status 2>&1 | grep 'connect ECONNREFUSED'
```
bingen: I'm now realizing that I had this for testing, but I would remove this one and keep the commented out one. This way we don't see that error message in the output while waiting for DB, which is normal but could be misleading (plus we already have the "Waiting for DB..." one above).

packages/server/scripts/db-setup.sh
```diff
@@ -15,7 +14,6 @@ done
 echo "DB is up!"
 
 # Check DB creation
-#npx sequelize db:migrate:status 2>&1 | grep 'database .* does not exist' > /dev/null
 npx sequelize db:migrate:status 2>&1 | grep 'database .* does not exist'
```
bingen: Same as above.

.github/workflows/mainnet.yml
facuspagnuolo: to see if you catch it haha, congrats 😄

docker-compose.yml
facuspagnuolo: Yea! Actually `nodemon` should avoid building the package, it is set up to work with `babel-node` directly

docker-compose.yml
facuspagnuolo: React-script uses it for hot-reload, we are using it to start the dashboard app :)

Dockerfile
facuspagnuolo: Actually no, based on the documentation it seems like `ADD` supports URLs, but we are not using that. I can change it, no strong opinion

promaty: ADD also can do some "magic" like automatically extracting zip files, so better to use COPY to avoid surprises

docker-compose.yml
```diff
@@ -8,20 +8,41 @@ services:
...
       - court_server
+    volumes:
+      - ./packages/services/bin:/court-backend/packages/services/bin
+      - ./packages/services/src:/court-backend/packages/services/src
+    env_file:
+      - .env
+
+  court_backoffice:
+    build: .
+    image: aragon/court-backoffice
```
promaty: Just wanted to say that in this case you are not creating another image, just adding a tag to the same image as before.
So you end up with the same Dockerfile image with 3 tags:
- aragon/court-server, services and backoffice.
I would just call this court-backend in all 3 cases to avoid confusion.
(The only difference is in how you execute it, i.e. the `command` line

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# copy root package and lerna json files
+COPY ./package.json /court-backend/package.json
+COPY ./lerna.json /court-backend/lerna.json
+
+# copy app package json
+RUN mkdir -p /packages/app
+COPY ./packages/app/package.json /court-backend/packages/app/package.json
+
+# copy cli package json
+RUN mkdir -p /packages/cli
```
sohkai, line 15: If we don't need the CLI in production (is it useful?), we could likely ignore this to shave off a bit of npm install time.

If lerna complains (I don't think it will if we just don't copy it), we can use a specific scope to ignore the cli package.

docker-compose.yml
```diff
@@ -8,20 +8,41 @@ services:
...
+
+  court_backoffice:
+    build: .
+    image: aragon/court-backend
+    command: npm run start:app
+    depends_on:
+      - court_server
+    ports:
+      - "3000:3000"
+      - "35729:35729"
```
sohkai, line 43: Could we add a comment about the hot reloading port?

PR #95, opened 2020-03-19
Server: Polish main entry point

facuspagnuolo:
**MERGE AFTER #94**

### DIFF
--- a/packages/server/src/index.js
+++ b/packages/server/src/index.js
@@ -10,23 +10,15 @@ import morgan from 'morgan'
 import { createServer } from '@promster/server'
 import { createMiddleware, signalIsUp } from '@promster/express'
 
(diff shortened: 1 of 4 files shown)


packages/server/src/index.js
```diff
@@ -10,23 +10,15 @@ import morgan from 'morgan'
...
 
-import errorHandler from './helpers/error-handler'
 import routes from './routes'
-import db from './models'
+import errorHandler from './helpers/error-handler'
+import checkConnection from './database/check-connection'
 
+// Load env variables and check DB connection
 dotenv.config()
+checkConnection().then(console.log)
```
sohkai, line 19: Wondering, would it be better to only start the express app if we were able to connect to the db?

PR #111, opened 2020-03-25
Shared: Add monitoring setup

promaty:
Checkout the monitoring setup, I've added some details to readme, you need a loki docker plugin before running docker compose:
```
docker plugin install  grafana/loki-docker-driver:latest --alias loki --grant-all-permissions
docker-compose up --build -d
```
This is only available for development, but I made the dashboard compatible with kubernetes so shouldn't take long to have a setup in the cluster, hopefully less than a day.


monitoring/grafana/provisioning/dashboards/dashbaords.yaml
```diff
@@ -0,0 +1,8 @@
+# This config loads all JSON grafana dashboards from the current directory
```
sohkai, line 1: Nitpick: the filename has a typo

promaty: eagle eye :D

promaty: Added to rest api PR: https://github.com/aragonone/court-backend/pull/132/commits/8828a5a1d12cc67cdb687c6b0cd77f1c8572750a

PR #113, opened 2020-03-27
Service: Improve monitor keeper

### LINKED ISSUE #85: Service: Implement notification for low ETH
To `monitor-address` service, for the account that is running the services.

facuspagnuolo:
Fixes #85 

Reducing the number of requests to Infura, we can use Etherscan for this one

### DIFF
--- a/packages/server/src/database/migrations/20200326212213-create-keeper-suspicious-transactions.js
+++ b/packages/server/src/database/migrations/20200326212213-create-keeper-suspicious-transactions.js
@@ -1,0 +1,34 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) =>
+    queryInterface.createTable('KeeperSuspiciousTransactions',
+      {
+        id: {
+          allowNull: false,
+          autoIncrement: true,
+          primaryKey: true,
+          type: Sequelize.INTEGER
+        },
+        blockNumber: {
+          type: Sequelize.STRING,
+          allowNull: false,
+          unique: false,
+        },
+        txHash: {
+          type: Sequelize.STRING,
+          allowNull: true,
+          unique: true,
+        },
+        createdAt: {
+          allowNull: false,
+          type: Sequelize.DATE
+        },
+        updatedAt: {
+          allowNull: false,
+          type: Sequelize.DATE
+        }
+      }
+    ),
+
+  down: (queryInterface, Sequelize) =>
+    queryInterface.dropTable('KeeperSuspiciousTransactions')
+}
--- a/packages/services/src/models/Etherscan.js
+++ b/packages/services/src/models/Etherscan.js
@@ -1,0 +1,41 @@
+import axios from 'axios'
+import { bn } from '@aragonone/court-backend-shared/helpers/numbers'
+
+const BASE_URL = 'https://<network>.etherscan.io/api?'
+
+export default class Etherscan {
+  constructor(network, key) {
+    this.key = key
+    this.url = this._getBaseUrl(network)
+  }
+
+  async getBalance(address) {
+    const path = `${this.url}module=account&action=balance&address=${address}&tag=latest`
+    const response = await this._get(path)
+    if (response.status !== '1' || !response.result) throw Error(`Received error from Etherscan: ${JSON.stringify(response)}`)
+    return bn(response.result)
+  }
+
+  async getTransactionsFrom(address, fromBlock) {
+    const path = `${this.url}module=account&action=txlist&address=${address}&startblock=${fromBlock}&sort=desc`
+    const response = await this._get(path)
+
+    if ((response.status !== '1' && response.message !== 'No transactions found') || !response.result)
+      throw Error(`Received error from Etherscan: ${JSON.stringify(response)}`)
+
+    return response.result
+      .filter(({ from }) => from === address)
+      .map(({ blockNumber, hash, to, value }) => ({ blockNumber, hash, value, to: to.toLowerCase() }))
+  }
+
+  async _get(url) {
+    const { data } = await axios.get(`${url}&apikey=${this.key}`)
+    return data
+  }
+
+  _getBaseUrl(network) {
+    if (network === 'rinkeby' || network === 'ropsten') return BASE_URL.replace('<network>', network)
+    else if (network === 'staging' || network === 'usability') return BASE_URL.replace('<network>', 'rinkeby')
+    return BASE_URL.replace('<network>', 'api')
+  }
+}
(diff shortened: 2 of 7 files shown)


facuspagnuolo:
Merging to test it, please review @sohkai

packages/services/src/models/Etherscan.js
```diff
@@ -0,0 +1,41 @@
...
+    const response = await this._get(path)
+    if (response.status !== '1' || !response.result) throw Error(`Received error from Etherscan: ${JSON.stringify(response)}`)
+    return bn(response.result)
+  }
+
+  async getTransactionsFrom(address, fromBlock) {
+    const path = `${this.url}module=account&action=txlist&address=${address}&startblock=${fromBlock}&sort=desc`
+    const response = await this._get(path)
+
+    if ((response.status !== '1' && response.message !== 'No transactions found') || !response.result)
```
sohkai, line 23: This status code check should probably be for `'200'`? Or does Etherscan return `1` as an HTTP status code?

packages/services/src/models/Etherscan.js
```diff
@@ -0,0 +1,41 @@
...
+      .map(({ blockNumber, hash, to, value }) => ({ blockNumber, hash, value, to: to.toLowerCase() }))
+  }
+
+  async _get(url) {
+    const { data } = await axios.get(`${url}&apikey=${this.key}`)
+    return data
+  }
+
+  _getBaseUrl(network) {
+    if (network === 'rinkeby' || network === 'ropsten') return BASE_URL.replace('<network>', network)
```
sohkai, line 37: I think for test networks, the api is `api-<network>`

packages/services/src/workers/monitor-keeper.js
```diff
@@ -1,88 +1,90 @@
...
+const { Admin, KeeperSuspiciousTransaction } = Models
 
 export default async function (logger) {
-  const court = await Network.getCourt()
-  const { environment } = court
-  const keeper = await getKeeperAddress(environment)
-  const courtAddresses = await getWhitelistedAddresses(court)
+  const { environment: { network } } = Network
+  const etherscan = new Etherscan(network, process.env.ETHERSCAN_API_KEY)
+  const keeper = getWalletFromPk(process.env.PRIVATE_KEY).getAddressString()
```
sohkai, line 18: Sanity checking—what happens here if we don't pass in a private key (e.g. during local development)?

packages/services/src/workers/monitor-keeper.js
```diff
@@ -1,88 +1,90 @@
...
+  if (transactions.length === 0) return logger.info('No transactions found')
+
   try {
-    const currentBlockNumber = await environment.getLastBlockNumber()
-    logger.info(`Checking transactions for keeper address ${keeper} from block ${lastCheckedBlockNumber + 1} to ${currentBlockNumber}`)
-    for (let blockNumber = lastCheckedBlockNumber + 1; blockNumber <= currentBlockNumber; blockNumber++) {
-      await checkTransactions(logger, environment, courtAddresses, blockNumber, keeper)
+    for (const transaction of transactions) {
+      const { hash, to, value, blockNumber } = transaction
+      logger.info(`Found transaction ${hash} on block ${blockNumber}`)
```
sohkai, line 35: For the logging, maybe it would make sense to put this in an else clause after the if, so we avoid seeing both this and then the "suspicious" entry in the logs.

packages/services/src/workers/monitor-keeper.js
```diff
@@ -1,88 +1,90 @@
...
   }
-}
 
-async function checkTransactions(logger, environment, courtAddresses, blockNumber, address) {
-  logger.info(`Checking transactions of block ${blockNumber}`)
-  const block = await environment.getBlock(blockNumber)
+  const { blockNumber } = transactions[0]
+  const last = await KeeperSuspiciousTransaction.last()
+  !last || !!last.txHash ? await KeeperSuspiciousTransaction.create({ blockNumber }) : await last.udpate({ blockNumber })
+  logger.success(`Checked transactions until block ${lastInspectedBlockNumber} successfully`)
```
sohkai, line 49: Would reword to "Successfully checked transactions until..."

packages/server/src/database/migrations/20200326212213-create-keeper-suspicious-transactions.js
```diff
@@ -0,0 +1,34 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) =>
+    queryInterface.createTable('KeeperSuspiciousTransactions',
+      {
+        id: {
+          allowNull: false,
+          autoIncrement: true,
+          primaryKey: true,
+          type: Sequelize.INTEGER
+        },
```
sohkai, line 10: I wonder if we should make this table be based on the keeper address—as it is right now, there's only one row anyway, right?

packages/services/src/workers/monitor-keeper.js
facuspagnuolo: It fails :)

packages/services/src/models/Etherscan.js
facuspagnuolo: Note that this `status` is part of the response data, not the HTTP header. Try these queries:
1. [successful](http://api.etherscan.io/api?module=account&action=txlist&address=0xde0b295669a9fd93d5f28d9ec85e40f4cb697bae&startblock=0&endblock=99999999&sort=asc&apikey=YourApiKeyToken): you will see that `status` is `1` here
2. [failure](http://api.etherscan.io/api?module=account&action=txlist&address=0xfe0b295669a9fd93d5f28d9ec85e40f4cb697bae&startblock=0&endblock=99999999&sort=asc&apikey=YourApiKeyToken): you will see that `status` is `0` here

They also provide an `isError` field, but as you can see this does variable does not appear in the second scenario.

packages/server/src/database/migrations/20200326212213-create-keeper-suspicious-transactions.js
facuspagnuolo: Do you mean storing entries per keeper address to support many of them?

sohkai: Yes, I think that's what I meant before, in case we had multiple addresses or swapped the current one out.

facuspagnuolo: Cool! I'll create an issue for that

PR #118, opened 2020-03-30
Chore: Add CI/CD for Grafana dashboard

promaty:


.dockerignore
```diff
@@ -1 +1,2 @@
 .github
+monitoring
```
facuspagnuolo: If we ignore this dir, will docker know when the grafana dashboard json changes then?

monitoring/prometheus/prometheus.yml
```diff
@@ -1,6 +1,6 @@
 # This config is required for Prometheus to scrape other containers
 scrape_configs:
-- job_name: court-backend
+- job_name: default/court-backend
```
facuspagnuolo: just curious, but why adding `default` here?

monitoring/grafana/provisioning/dashboards/court-backend.json
```diff
@@ -253,5 +246,5 @@
   "variables": {
     "list": []
   },
-  "version": 1
+  "version": 2
```
facuspagnuolo: should we update the version number every time we change this file? grafana wouldn't update it otherwise?

.dockerignore
promaty: Docker doesn't need to know, dashboard is deployed separately
(mounted as docker-compose volume locally and deployed as k8s object in the cluster)

monitoring/prometheus/prometheus.yml
promaty: When scraping in k8s all resources have a namespace, which is "default" by default :)
This makes the dashboard parsing the same on localhost and in k8s

monitoring/grafana/provisioning/dashboards/court-backend.json
promaty: When you edit the dashboard and click save -> export JSON this number is generated automatically.

facuspagnuolo:
Thanks for the clarifications @promaty !

PR #120, opened 2020-03-31
Server: Implement basic metrics

### LINKED ISSUE #37: Server: Add Prometheus metrics
We could some simple metrics to track:
- \# of total requests
- \# of requests per endpoint
- \# of total errors
- \# of errors per HTTP code
- \# of DB queries
- \# of DB errors

facuspagnuolo:
Fixes #37 

Quick proof of concept of how this could be implemented

image

image


facuspagnuolo:
@promaty note that we were also submitting information about the content of each error, I think that was the whole point of it. Apart from that, are the HTTP requests per path also covered by promster by default?

promaty:
> @promaty note that we were also submitting information about the content of each error, I think that was the whole point of it. Apart from that, are the HTTP requests per path also covered by promster by default?

Yes the path is covered in total requests.

I missed the part where you added the content of HTTP error. I don't think we can use complex strings and stringified objects as metrics label, they have to be simple.

I think that the best way to do this is just plot default HTTP requests per path/code and have more detailed error strings in our logs.

packages/server/src/controllers/AdminsController.js
```diff
@@ -9,8 +9,10 @@ export default {
   async login(request, response, next) {
     try {
       const errors = await AdminValidator.validateForLogin(request.body)
-      if (errors.length > 0) return response.status(400).send({ errors: errors })
-
+      if (errors.length > 0) {
+        response.status(400).send({ errors: errors })
+        next()
```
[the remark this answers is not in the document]
facuspagnuolo: We actually did that, look at the last commit

facuspagnuolo:
Then I think this PR doesn't add any server metric at all, we are just relying on promster ones.

promaty:
Yes but we have error handler and metrics reporter objects so that's useful. We can add some for DB errors for example.

promaty:
Maybe we can also expand morgan logs object to print error messages.

facuspagnuolo:
@promaty added the last tweak to track DB errors. I also checked the morgan docs for logging requests, but I think the `dev` option we are currently using is quite useful for now. Feel free to merge if it looks good to you

packages/server/src/errors/http-error.js
```diff
@@ -0,0 +1,15 @@
+export default class HttpError extends Error {
+  static _400(content) {
```
sohkai, line 2: The `_` prefix is new to me—I think we can just use numbers as method names?

But otherwise, I wonder what you think about naming these errors based on their status codes, e.g. naming them similar to (and perhaps also using the constants from) [http-status-codes](https://www.npmjs.com/package/http-status-codes)?

packages/server/src/helpers/metrics-reporter.js
```diff
@@ -0,0 +1,35 @@
...
+    { name: 'errors', help: 'DB errors' },
+  ],
+}
+
+class MetricsReporter {
+  constructor(app) {
+    this._initializeCounterMetrics(app)
+  }
+
+  dbQuery() {
```
sohkai, line 13: Do we actually use this metric at the moment?

packages/server/src/errors/http-error.js
promaty: I think you cannot use a method/function that starts with a number.

I like the idea of using status codes, would be more readable.

packages/server/src/helpers/metrics-reporter.js
promaty: not yet, I think this was a contrived example just to implement the error/metric handling. All most important metrics currently are HTTP requests by path/code, which is handled by promster. 
I think metrics will be more useful in services so we can reuse this logic there.

PR #126, opened 2020-04-03
Server: Improve create reveal endpoint

facuspagnuolo:


packages/server/src/controllers/RevealsController.js
```diff
@@ -19,8 +20,12 @@ export default {
     const errors = await RevealsValidator.validateForCreate(params)
     if (errors.length > 0) throw HttpError._400({ errors })
 
+    const decodedVoteId = decodeVoteId(params.voteId)
     params.revealed = false
+    params.disputeId = decodedVoteId.disputeId.toString()
+    params.roundNumber = decodedVoteId.roundId.toString()
```
sohkai, line 26: Interesting... I wonder why missing this information wasn't failing before? We have the `notEmpty` validation set in the db schema.

facuspagnuolo: I think it was failing but we didn't know :/
I'll create an issue to add tests for these endpoints

PR #132, opened 2020-04-04
Server: dummy REST endpoints for notifications

promaty:
No complicated checks for address/email/signature formats, all endpoints return success.
The only error you can get is if there is no authenticated session.

You can test with the following:
`curl localhost:8000/users/testaddr/sessions -X POST -v`
Get the `Set-Cookie: connect.sid=<secret>` string
Append `-H "Cookie: connect.sid=<secret>"` to following requests

I think this is good enough to start developing frontend


packages/server/src/controllers/UsersController.js
```diff
@@ -4,7 +4,135 @@ import UsersValidator from '../validators/UsersValidator'
...
+
 export default {
+  async details(req, res) {
+    let body = {
+      "emailExists": false,
+      "emailVerified": false,
+      "addressVerified": false,
+      "notificationsEnabled": false
+    }
+    Object.assign(body, Users[req.params.address])
```
facuspagnuolo, line 21: Any reason for not doing simply 
```js
Users[req.params.address] = body
```

packages/server/src/controllers/UsersController.js
```diff
@@ -4,7 +4,135 @@ import UsersValidator from '../validators/UsersValidator'
...
+  sessions: {
+    async create(req, res) {
+      if (!(req.params.address in Users)) {
+        Users[req.params.address] = {}
+      }
+      Users[req.params.address]['addressVerified'] = true
+      if (!(req.params.address in UserSessions)) {
+        UserSessions[req.params.address] = {}
+      }
+      UserSessions[req.params.address][req.session.id] = true
```
facuspagnuolo, line 35: To avoid repeating `req.params.address` every time you can extract it at the beginning of each function as follows:
```js
const { params: { address } }  = req
```

packages/server/src/controllers/UsersController.js
```diff
@@ -4,7 +4,135 @@ import UsersValidator from '../validators/UsersValidator'
...
+
+    async authenticate(req,res,next) {
+      if (
+        !(req.params.address in UserSessions) ||
+        !(req.session.id in UserSessions[req.params.address])
+      ) {
+        // exception for new emails from anj
+        if (req.path.endsWith("/email") && !(req.params.address in Users)) {
+          return next()
+        }
```
facuspagnuolo, line 50: I think skipping particular paths here is a bit weird, we can do something like the `asyncMiddleware` for authenticated routes, see my example [here](https://github.com/aragonone/court-backend/pull/133/files?file-filters%5B%5D=.js&file-filters%5B%5D=.sample#diff-d8751b0865f851a3e04a955c3c19aa33R46-R49)

packages/server/src/controllers/UsersController.js
```diff
@@ -4,7 +4,135 @@ import UsersValidator from '../validators/UsersValidator'
...
+let UserSessions = { }
+
+
 export default {
+  async details(req, res) {
+    let body = {
+      "emailExists": false,
+      "emailVerified": false,
+      "addressVerified": false,
+      "notificationsEnabled": false
```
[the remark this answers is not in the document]
facuspagnuolo: Shouldn't we read these attributes from `Users[address]`?

packages/server/src/controllers/UsersController.js
```diff
@@ -4,7 +4,135 @@ import UsersValidator from '../validators/UsersValidator'
...
+      res.send(body)
+    },
+
+    async change(req, res) {
+      UserEmails[req.params.address] = req.body.email
+      if (!(req.params.address in Users)) {
+        Users[req.params.address] = {}
+      }
+      Users[req.params.address]['emailExists'] = true
+      Users[req.params.address]['notificationsEnabled'] = true
```
facuspagnuolo, line 89: We shouldn't change these here right?

packages/server/src/controllers/UsersController.js
```diff
@@ -4,7 +4,135 @@ import UsersValidator from '../validators/UsersValidator'
...
+        "deleted": true
+      }
+      res.send(body)
+    },
+  },
+
+
+  notifications: { 
+    async change(req, res) {
+      Users[req.params.address]['notificationsEnabled'] = req.body.enabled
```
facuspagnuolo, line 127: Note that this should change the notification settings, not the opted-out flag

packages/server/src/controllers/UsersController.js
```diff
@@ -4,7 +4,135 @@ import UsersValidator from '../validators/UsersValidator'
...
+        "sent": true
+      }
+      res.send(body)
+    },
+    
+    async delete(req, res) {
+      delete UserEmails[req.params.address]
+      Users[req.params.address]['emailExists'] = false
+      Users[req.params.address]['emailVerified'] = false
+      Users[req.params.address]['notificationsEnabled'] = false
```
facuspagnuolo, line 116: I think we should delete the sessions as well

packages/server/src/controllers/UsersController.js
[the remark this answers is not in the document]
promaty: Good Nit, removed quotes in new commit.

> Shouldn't we read these attributes from `Users[address]`?
Yes these will be a default return from the model, but I don't have it for now.

packages/server/src/controllers/UsersController.js
promaty: Actually it's the other way around :)
```js
body = Users[req.params.address]
```
Some keys ins `Users[req.params.address]` may not exist so it will be an empty return instead of `false`

packages/server/src/controllers/UsersController.js
promaty: oh nice thanks

packages/server/src/controllers/UsersController.js
promaty: correct, I should put this in the default model value

packages/server/src/controllers/UsersController.js
promaty: Hmm not sure, there are two separate buttons for log out and delete email, if we automatically logout when deleting email it will look like a bug.

promaty:
> * I don't see the need for overwriting the `emailExists`, `emailVerified`, `addressVerified` and `notificationsEnabled` keys deliberately
I think only `notificationsEnabled` was overwritten out of place, others only change when hitting relevant API

> * I still think `notificationsEnabled` does not reflect what we are trying to communicate, I think this should be `optedOut`, `accountDisabled`, or similar. Otherwise, I find it misleading with someone that hasn't opted out but has turned the notifications off for a while.

I changed it to `notificationsDisabled`, I think this will be ok.
- `false` by default
- `true` will reflect when user clicks notifications slider, clicking opting out
- We will be consistent with other terms (noun + action) rather than something like optedout

packages/server/src/controllers/UsersController.js
promaty: Yes I agree but we need to check this special case somehow wherever it is.
I was thinking why do we even allow this anonymous situation? Let's just ask every new subscriber to authenticate and verify same as for the client (we will only treat previously subscribed jurors in a special way).

packages/server/src/routes/index.js
```diff
@@ -5,6 +5,27 @@ import { admins, users, reveals } from '../controllers'
...
+
+  // manage user emails
+  app.get(    '/users/:address/email',                asyncMiddleware(users.email.get))
+  app.put(    '/users/:address/email',                asyncMiddleware(users.email.change))
+  app.post(   '/users/:address/email[:]verify',       asyncMiddleware(users.email.verify))
+  app.post(   '/users/:address/email[:]send',         asyncMiddleware(users.email.send))
+  app.delete( '/users/:address/email',                asyncMiddleware(users.email.delete))
+
+  // set notifications
+  app.put(    '/users/:address/notifications',        asyncMiddleware(users.notifications.change))
```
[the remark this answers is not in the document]
sohkai: Haha... once we apply prettier to this repo... I don't think we'll be able to keep this type of formatting. But let's see what we can do :).

packages/server/src/controllers/UsersController.js
```diff
@@ -4,7 +4,135 @@ import UsersValidator from '../validators/UsersValidator'
...
 
+// temp dummy models
+let Users = { }
+let UserEmails = { }
+let UserSessions = { }
+
+
 export default {
+  async details(req, res) {
+    let body = {
```
sohkai, line 16: Generally prefer `const` over `let`, unless you know the variable is going to change :)

packages/server/src/controllers/UsersController.js
```diff
@@ -4,7 +4,145 @@ import UsersValidator from '../validators/UsersValidator'
...
 export default {
+  async details(req, res) {
+    const { params: { address } } = req
+    let body = {
+      emailExists: false,
+      emailVerified: false,
+      addressVerified: false,
+      notificationsDisabled: false
+    }
+    Object.assign(body, Users[address])  // Object.assign acts as a merge to avoid empty keys
```
sohkai, line 22: `Object.assign()` has usually been superceded by the object-rest-spread operator instead:

```
Object.assign(body, Users[address])
```

is equivalent to the usually more preferred version:

```
const body = {
  key: 'value',
  ...Users[address],
}
```

This may be a useful time machine reference for syntax like this: https://github.com/Khan/style-guides/blob/master/style/javascript.md#dont-use-underscore

packages/server/src/controllers/UsersController.js
```diff
@@ -4,7 +4,145 @@ import UsersValidator from '../validators/UsersValidator'
...
+    res.send(body)
+  },
+
+
+  sessions: {
+    async create(req, res) {
+      const { params: { address } } = req
+      if (!(address in Users)) {
```
sohkai, line 30: _Usually_ style guides will prefer you to avoid `in` because it has loose semantics (more verbose than just property access, less specific than just `hasOwnProperty()`).

This is a bit of a nitpick now because this operation would involve a DB query later and will get replaced, but just in case we keep some of the code for later :).

packages/server/src/controllers/UsersController.js
promaty: oh nice will update

packages/server/src/controllers/UsersController.js
promaty: Removed this part, let's just reuse your old `POST` `/users` endpoint for anonymous requests.

PR #133, opened 2020-04-06
Server: Use session cookies for admin routes

### LINKED ISSUE #121: Server: Use cookies + CSRF token instead of JWTs

### LINKED ISSUE #122: Server: Send 404 instead of 403 for unknown routes

facuspagnuolo:
**MERGE AFTER #132**

Fixes #121 
Fixes #122


.env.sample
```diff
@@ -24,10 +26,10 @@ REACT_APP_NETWORK=rpc
 REACT_APP_SERVER_URL=http://localhost:8000
 
 # Web3 stuff config
```
sohkai, line 28: Haha maybe we can split these options into "Web3 connection config" and "Web3 signing config"

.env.sample
```diff
@@ -24,10 +26,10 @@ REACT_APP_NETWORK=rpc
...
-PRIVATE_KEY=
 NETWORK=rpc
-RPC=http://localhost:8545
-COURT_ADDRESS=0xD833215cBcc3f914bD1C9ece3EE7BF8B14f841bb
-GAS_PRICE=15000000001
 GAS=7000000
+GAS_PRICE=15000000001
 WEB3_POLLING_INTERVAL=10000
+COURT_ADDRESS=0xD833215cBcc3f914bD1C9ece3EE7BF8B14f841bb
```
sohkai, line 33: Similar nitpick, but this seems like it could be at the top of the list or etc. since it's very fundamental.

packages/app/src/store/local.js
```diff
@@ -0,0 +1,37 @@
+import Network from '../web3/Network'
```
sohkai, line 1: Is saving the token in local storage useful if it's already persisted as a secure token? One of the nice things of secure tokens is that it's protected from JS :)

packages/server/README.md
```diff
@@ -54,8 +54,9 @@ All the provided endpoints are `Content-Type: application/json`
 
 - URL: /users
 - Method: GET
+- Header:
+  - Cookie: `aragonCourtSID=<SID>`
```
sohkai, line 58:
```suggestion
  - Cookie: `aragonCourtSessionID=<SID>`
```

Might be clearer

packages/server/src/routes/index.js
```diff
@@ -25,23 +27,28 @@ export default app => {
   app.put(    '/users/:address/notifications',        asyncMiddleware(users.notifications.change))
 
 
-  // other endpoints
-  app.post('/login', asyncMiddleware(admins.login))
+  /*********** Old routes ***********/
```
sohkai, line 30: Should we just remove these old routes? And do we want to consider `/reveals` an old route?

packages/server/src/helpers/cors-middleware.js
```diff
@@ -1,8 +1,10 @@
 import cors from 'cors'
 
 const CORST_OPTIONS = {
+  credentials: true,
   origin: function (origin, callback) {
     const whitelist = process.env.CORS_WHITELIST.split(',')
+    console.log(origin)
```
sohkai, line 7: Should we remove this once we're done testing?

packages/server/src/models/Session.js
```diff
@@ -0,0 +1,49 @@
...
+        }
+      },
+      modelId: {
+        type: DataTypes.INTEGER,
+        allowNull: false,
+        validate: {
+          notEmpty: true,
+        }
+      },
+      modelType: {
```
sohkai, line 19: Not sure I understand the purpose of `modelId` and `modelType`—how are we using them?

packages/server/src/routes/index.js
facuspagnuolo: Well, we are currently using them. I mean "old" in terms of routes we haven't sanitized yet with the RESTful rules we applied to the new ones, would love to update these in the future too

packages/server/src/models/Session.js
facuspagnuolo: Currently, we only have admin sessions, but we will have also users sessions. The idea is that we can track them in the same table but having a polymorphic relation for each of them

packages/app/src/store/local.js
facuspagnuolo: Not sure I follow this comment, I tried to access the SID from the cookies from JS but I couldn't, that's why I stored them in local storage

packages/server/src/routes/index.js
facuspagnuolo: Left an issue to tackle that here https://github.com/aragonone/court-backend/issues/135 :)

packages/server/src/models/Session.js
sohkai: Wondering—would an `admin` boolean column suffice or would it be awkward?

facuspagnuolo: I could work, will check if there is a way to define polymorphic relations with a boolean

packages/app/src/web3/Server.js
```diff
@@ -1,25 +1,24 @@
 import axios from 'axios'
```
promaty, line 1: can't we just use fetch for this stuff?

packages/app/src/store/local.js
promaty: I thought the main point of using cookies was that you cannot access it from js? SID should be set automatically by the browser when doing HTTP requests and then server decides from that header what your permissions are. You should never see SID on the client.

facuspagnuolo: Yes, it is actually doing that automatically, this is simply to know if the admin has logged in or not to show some options in the client, basically after a successful login, I set a local storage variable to identify that state

packages/server/src/routes/authenticate-admin.js
```diff
@@ -0,0 +1,21 @@
...
+
+const { Admin } = Models
+
+const authenticate = route => async (request, response, next) => {
+  const { session, cookies } = request
+
+  if (!session.modelId || !session.modelType) {
+    // Check if user's cookie is still saved in browser and user is not set, then automatically log the user out
+    // This usually happens when you stop your express server after login, your cookie still remains saved in the browser
+    if (cookies && cookies.aragonCourtSessionID) response.clearCookie('aragonCourtSessionID')
```
promaty, line 13: I don't think this changes much. If there is no valid cookie in the backend, the server will just ignore it and automatically respond with a new `set-cookie` header

facuspagnuolo: It would be like forcing that header, not sure if the server does that automatically, but if it does then we can remove this line

packages/app/src/web3/Server.js
facuspagnuolo: No, we can create an issue to stop using axios

packages/server/src/routes/authenticate-admin.js
promaty: that's quite standard for session cookies, browser always syncs with whatever is inside `set-cookie`

promaty: Also it's going to mess up the request if for example you want to authenticate (create cookie) but you also deleting a cookie in the same request. better to remove this.

packages/app/src/web3/Server.js
promaty: So you are using it in other places as well? Why is that?

facuspagnuolo: No particular reason, we can simply use fetch. We are using it for the monitoring service as well

facuspagnuolo: Created [this issue](https://github.com/aragonone/court-backend/issues/139) to tackle that

PR #136, opened 2020-04-06
Server: Sanitize addresses for all endpoints

facuspagnuolo:


packages/server/src/helpers/sanitize-address-middleware.js
```diff
@@ -0,0 +1,14 @@
...
+
+const sanitizeAddresses = object => {
+  Object.keys(object).forEach(key => {
+    if (!!object[key] && typeof object[key] === 'object') sanitizeAddresses(object[key])
+    else if(isAddress(object[key])) object[key] = object[key].toLowerCase()
+  })
+}
+
+export default () => (req, res, next) => {
+  sanitizeAddresses(req.query)
```
sohkai, line 11: Ah, this is quite a cute middleware trick! Never thought of doing it like this, but it's a good way to dedupe the logic!

packages/server/src/helpers/sanitize-address-middleware.js
```diff
@@ -0,0 +1,14 @@
+import { isAddress } from 'web3-utils'
+
+const sanitizeAddresses = object => {
+  Object.keys(object).forEach(key => {
```
sohkai, line 4: We could likely simplify this via `Object.entries()`, which will give us both the keys and values

packages/server/src/helpers/sanitize-address-middleware.js
```diff
@@ -0,0 +1,14 @@
...
+    if (!!object[key] && typeof object[key] === 'object') sanitizeAddresses(object[key])
+    else if(isAddress(object[key])) object[key] = object[key].toLowerCase()
```
sohkai, line 6: I would do a stricter check here—let's transform to checksum and error if that fails before passing it through as all lowercased.

sohkai: Ah, actually I think `isAdress()` checks for the checksum already right? In that case, ignore me :).

packages/server/src/helpers/sanitize-address-middleware.js
facuspagnuolo: definitely!

packages/server/src/helpers/sanitize-address-middleware.js
facuspagnuolo: yes it does :)

PR #144, opened 2020-04-22
CI/CD: moved all ci/cd logic into shared scripts

promaty:
Looks much cleaner now :)


promaty:
Sorry messed up the branches, rebased should be good now.

PR #145, opened 2020-04-27
Server: Implemented Knex/Objection storage

promaty:


promaty:
Let's keep this open while I migrate other tables, should only take a day or so.

promaty:
Users/Sessions/Emails are fully migrated to knex. I also have automated test for the POST /users endpoint. Now just Admins and Reveals.

promaty:
I did what I could, but left Admins and Reveals refactoring for you.
Few notes:
- Knex migrations run correctly both on existing Sequelize objects and on fresh DB (we can remove sequelize stuff from `db-setup.sh` without problems)
- Users controller is completely refactored for the new objection models
- I have included Admins and Reveals models as well so you can use them in migration
- Currently GET /users endpoint breaks in admin panel, because user results are returned in a different format, I have left some comments.

packages/server/src/controllers/UsersController.js
```diff
@@ -1,172 +1,157 @@
...
+
+
+  async create(req, res) {
+    const params = req.body
+    const errors = await UsersValidator.validateForCreate(params)
+    if (errors.length > 0) throw HttpError._400({ errors })
+    const { email, address } = params
+    let userEmail = await UserEmails.query().findOne({email})
+    if (!userEmail) userEmail= await UserEmails.query().insert({email})
+    await userEmail.$relatedQuery('users').insert({address})
```
facuspagnuolo, line 27: This is another candidate:

```js
async associateAddress(address) {
   await this.$relatedQuery('users').insert({address})
}
```

packages/server/src/database/knex/migrations/20200425004140_create-admins.js
```diff
@@ -0,0 +1,23 @@
...
+  return knex.schema.hasTable('Admins').then(function(sequelizeExists) {
+    if (sequelizeExists) {
+      return knex.schema.alterTable('Admins', function (table) {
+        table.datetime('createdAt').defaultTo(knex.fn.now()).alter()
+        table.datetime('updatedAt').defaultTo(knex.fn.now()).alter()
+      })
+    } else {
+      return knex.schema.createTable('Admins', function (table) {
+        table.increments('id')
+        table.string('email').unique()
```
facuspagnuolo, line 12: It would be useful to have an index here, not sure if `unique` does it by default

packages/server/src/database/knex/migrations/20200425004137_create-users.js
```diff
@@ -0,0 +1,29 @@
...
+        table.renameColumn('userId', 'userEmailId')
+        table.dropForeign('userEmailId', 'UserAddresses_userId_fkey') // need to recreate constraint with set null on delete
+        table.foreign('userEmailId').references('UserEmails.id').onDelete('SET NULL')
+        table.datetime('createdAt').defaultTo(knex.fn.now()).alter()
+        table.datetime('updatedAt').defaultTo(knex.fn.now()).alter()
+      })
+    } else {
+      return knex.schema.createTable('Users', function (table) {
+        table.increments('id')
+        table.string('address').unique()
```
facuspagnuolo, line 16: It would be useful to have an index here, not sure if `unique` does it by default

packages/server/src/database/knex/migrations/20200425004150_create-sessions.js
```diff
@@ -0,0 +1,18 @@
+
+exports.up = function(knex) {
+  return knex.schema.dropTableIfExists('Sessions').createTable('Sessions', function (table) {
+    table.increments('id')
+    table.string('sid').unique()
+    table.jsonb('data')
+    table.integer('adminId')
+    table.foreign('adminId').references('Admins.id').onDelete('CASCADE')
+    table.integer('userId')
+    table.foreign('userId').references('Users.id').onDelete('CASCADE')
```
facuspagnuolo, line 10: Just curious, this work as nullable foreign keys right?

packages/server/src/database/knex/seeds/insert-admin.js
```diff
@@ -0,0 +1,15 @@
+require('dotenv').config()
+const bcrypt = require('bcryptjs')
+const ADMIN_EMAIL = process.env.ADMIN_EMAIL
+const ADMIN_PASSWORD = process.env.ADMIN_PASSWORD
```
facuspagnuolo, line 4: It seems like a good case for destructive assignments, you can do
```js
const { env: { ADMIN_EMAIL, ADMIN_PASSWORD } } = process 
```

packages/server/src/database/knex/seeds/insert-admin.js
```diff
@@ -0,0 +1,15 @@
...
+
+exports.seed = function(knex) {
+  return knex('Admins').where({'email': ADMIN_EMAIL}).then(function(rows) {
```
facuspagnuolo, line 7: Can we use the Admin model here?

packages/server/src/routes/authenticate-admin.js
```diff
@@ -6,9 +6,9 @@ const { Admin } = Models
 
 const authenticate = route => async (request, response, next) => {
   const { session } = request
-  if (!session.modelId || !session.modelType) throw HttpError._403({ errors: [{ status: 'Unauthorized' }] })
+  if (!session.adminId) throw HttpError._403({ errors: [{ status: 'Unauthorized' }] })
 
-  request.currentAdmin = await Admin.findByPk(session.modelId)
+  request.currentAdmin = await Admin.findByPk(session.adminId)
```
facuspagnuolo, line 11: Nice :)

packages/server/src/database/knex/migrations/20200425004140_create-admins.js
promaty: Yep unique is implemented by adding an index under the hood: https://www.postgresql.org/docs/9.4/ddl-constraints.html

packages/server/src/database/knex/migrations/20200425004150_create-sessions.js
promaty: hmm yes I didn't add not nullable for columns.

packages/server/src/database/knex/seeds/insert-admin.js
promaty: Ah thanks, I'm still not used to these :)

packages/server/src/controllers/UsersController.js
```diff
@@ -1,172 +1,157 @@
...
+  },
+
+
+  async create(req, res) {
+    const params = req.body
+    const errors = await UsersValidator.validateForCreate(params)
+    if (errors.length > 0) throw HttpError._400({ errors })
+    const { email, address } = params
+    let userEmail = await UserEmails.query().findOne({email})
+    if (!userEmail) userEmail= await UserEmails.query().insert({email})
```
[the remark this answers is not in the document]
promaty: Added

packages/server/src/database/knex/seeds/insert-admin.js
promaty: Looks like we can, added.

packages/server/src/controllers/UsersController.js
```diff
@@ -1,172 +1,157 @@
...
     const { params: { address } } = req
-    const body = {
-      emailExists: false,
-      emailVerified: false,
-      addressVerified: false,
-      notificationsDisabled: false,
-      ... Users[address]
-    }
-    res.send(body)
+    const user = await Users.query().findOne({address}).withGraphFetched('[email, emailVerificationToken, notificationSettings]')
```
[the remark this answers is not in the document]
promaty: Not sure about this since those relations are doing multiple joins. I prefer to be more explicit about what returns I am doing.

I think other abstraction suggestions are good, we can have a rule for creating abstract queries that include logic, I added a few.

packages/server/src/controllers/UsersController.js
promaty: I don't know, as long as the query is one line it doesn't bother me much. If you insist you can add some, see my other examples :)

packages/server/src/database/knex/migrations/20200425004137_create-users.js
promaty: It does

packages/server/src/database/knex/migrations/20200425004150_create-sessions.js
promaty: I've added nullable constraints in all migrations where necessary so it's more clear

promaty:
Addressed all comments. I will split controllers in the next PR for proper validators, let's focus on storage in this one.

promaty:
Managed to migrate all DB and testing stuff to ESM modules and babel. Now we can also use ES2020 stuff in tests and migration/seed files and there is no more `require` anywhere :)

packages/server/src/database/knex/migrations/20200425004150_create-sessions.js
promaty: Also added raw constraint below that either adminId or userId has to be not null.

packages/server/src/database/knex/migrations/20200425004137_create-users.js
promaty: Added some extra indexes in other places (e.g. foreign keys, expires at)

packages/server/src/database/knex/migrations/20200428185153_create-reveals.js
```diff
@@ -0,0 +1,30 @@
+
+export function up(knex) {
+  return knex.schema.hasTable('Reveals').then(function(sequelizeExists) {
+    if (sequelizeExists) {
+      return knex.schema.alterTable('Reveals', function (table) {
+        table.boolean('revealed').defaultTo(false).notNullable().alter()
```
promaty, line 6: @facuspagnuolo I've added default value false here, seems appropriate. Also we may want to add some extra indexes here not sure how you query this stuff.

packages/server/src/models/objection/index.js
```diff
@@ -0,0 +1,19 @@
+import Admin from './Admin'
```
sohkai, line 1: Note that you can do something like

```js
export { default as Admin } from './Admin'
```

Instead to not repeat yourself :).

PR #148, opened 2020-04-30
Server: Log reveal errors temporarily

facuspagnuolo:


facuspagnuolo:
> Question, why is your validator returning an array of errors? I think that we should always fail on the first error we encounter, that's also what we have in notification API spec.

Well, from a UX perspective you want to know all the things you need to fix instead of receiving them one by one

PR #152, opened 2020-05-05
Chore: Improve grafana dashboard

facuspagnuolo:


promaty:
Made some improvements to abstract away individual pods, now the dashbaoards should me more readable

PR #155, opened 2020-05-07
Services: notifications service - subscription reminder

promaty:


packages/services/package.json
```diff
@@ -22,6 +23,10 @@
     "@babel/core": "^7.7.7",
     "@babel/node": "^7.7.7",
     "@babel/preset-env": "^7.7.7",
-    "nodemon": "^2.0.2"
+    "chai": "^4.2.0",
+    "mocha": "^7.1.2",
+    "nodemon": "^2.0.2",
+    "sinon": "^9.0.2",
```
facuspagnuolo, line 29: be really careful about this library, maybe in a few years you will understand me :)

packages/services/src/models/notification-scanners/BaseModel.js
```diff
@@ -0,0 +1,34 @@
+const { env: { CLIENT_URL } } = process
+
+export default class BaseModel {
```
facuspagnuolo, line 3: Should we rename it to `NotificationScannerBaseModel` or similar?

packages/services/src/models/notification-scanners/BaseModel.js
```diff
@@ -0,0 +1,34 @@
...
+    return (
+      user?.email &&
+      !user?.notificationSetting?.notificationsDisabled &&
+      (user?.emailVerified || this._sendUnverified)
+    )
+  }
+  get _MINUTES() { return 60 * 1000 }
+  get _HOURS() { return 60 * this._MINUTES }
+  get _DAYS() { return 24 * this._HOURS }
+  get _CLIENT_URL() { return CLIENT_URL }
```
facuspagnuolo, line 15: we could probably define this as static methods

packages/services/src/models/notification-scanners/SubscriptionReminder.js
```diff
@@ -0,0 +1,34 @@
+import BaseModel from './BaseModel'
+import { User } from '@aragonone/court-backend-server/build/models/objection'
+
+class SubscriptionReminder extends BaseModel {
+  async scan() {
+    let notifications = []
+    const users = await User.query().where({emailVerified: false}).withGraphFetched('[email,emailVerificationToken]')
```
facuspagnuolo, line 7: we can defined a function in the base model like `getList` or `getUsers` for the queries

packages/services/src/models/notification-scanners/SubscriptionReminder.js
```diff
@@ -0,0 +1,34 @@
...
+    for (const user of users) {
+      /**
+       * Generate subscription reminders in two cases:
+       * 1) If user has an existing verification token that expired a day ago
+       * 2) If user has no token and address was never verified (previous subscriptions from https://anj.aragon.org/)
+       */
+      if (!user.email) continue
```
facuspagnuolo, line 14: Would the `continue` stop the loop execution if the condition is not met here?

packages/services/src/models/notification-scanners/SubscriptionReminder.js
```diff
@@ -0,0 +1,34 @@
...
+      if (!user.emailVerificationToken && user.addressVerified) continue
+      if (user.emailVerificationToken && user.emailVerificationToken.expiresAt > new Date(Date.now()-this._DAYS)) continue
```
facuspagnuolo, line 16: Maybe I'm confused, but haven't we been repeating this logic in other places? Is it something we could abstract in the user under a method like `user.hasSubscribed()`

packages/services/src/workers/notification-scanner.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+export default async function (logger) {
+  Object.keys(notificationScanners).forEach(async (model) => {
+    await tryRunScanner(logger, model)
+  })
+}
+
+export async function tryRunScanner(logger, model) {
+  const type = await UserNotificationType.findOrInsert({model})
+  const scanner = notificationScanners[model]
```
facuspagnuolo, line 16: we can add a validation here since this is probably going to fail if someone renames a file or adds a new type in the DB manually

packages/services/src/workers/notification-scanner.js
```diff
@@ -0,0 +1,33 @@
...
+  const { scannedAt } = type
+  const { scanPeriod } = scanner
+  if (scannedAt && scannedAt.getTime()+scanPeriod > Date.now()) return
```
facuspagnuolo, line 19: This is something we could abstract in the scanner, sth like `scanner.shouldScanNow()`

packages/services/src/workers/notification-scanner.js
```diff
@@ -0,0 +1,33 @@
...
+  const notifications = await scanner.scan()
+  for (const notification of notifications) {
+    const { address, details } = notification
+    const user = await User.query().findOne({address})
```
facuspagnuolo, line 23: I think it's time to have a method in the `BaseModel` to avoid repeating `.query().findOne()` in more places

packages/services/src/workers/notification-scanner.js
```diff
@@ -0,0 +1,33 @@
...
+    if (!await scanner.checkUser(user)) continue
+    await UserNotification.findOrInsert({
+      userId: user.id,
+      userNotificationTypeId: type.id,
+      details
+    })
+  }
+  await type.$query().update({scannedAt: new Date()})
```
facuspagnuolo, line 31:
```suggestion
  await type.$query().update({ scannedAt: new Date() })
```

packages/services/test/helpers/userNotificationTypeByModel.js
```diff
@@ -0,0 +1,5 @@
+import { UserNotificationType } from '@aragonone/court-backend-server/build/models/objection'
+
+export default function userNotificationTypeByModel(model) {
+  return UserNotificationType.query().findOne({model}).withGraphFetched('notifications')
```
facuspagnuolo, line 4: Would it be helpful to have it in the model? It seems like the fact of needing this for test is telling us something :)

packages/services/src/workers/notification-sender.js
```diff
@@ -0,0 +1,34 @@
...
+  for (const notification of notifications) {
+    await trySendNotification(logger, notification)
+  }
+}
+
+export async function trySendNotification(logger, notification) {
+  if (notification.sentAt != null) return
+  notification = await notification.$fetchGraph('[user.email, type]')
+  const { user, type: { model } } = notification
+  const scanner = notificationScanners[model]
```
facuspagnuolo, line 20: Same here for the scanner validation

packages/services/src/models/notification-scanners/BaseModel.js
```diff
@@ -0,0 +1,34 @@
+const { env: { CLIENT_URL } } = process
+
+export default class BaseModel {
+  async checkUser(user) {
+    user = await user.$fetchGraph('[email, notificationSetting]')
+    return (
+      user?.email &&
+      !user?.notificationSetting?.notificationsDisabled &&
+      (user?.emailVerified || this._sendUnverified)
```
facuspagnuolo, line 9: Something tells me this question belongs to the user :)

packages/services/src/models/notification-scanners/SubscriptionReminder.js
promaty: Yeah I just realised if I have to do comments it's probably too complicated :)
Logic is not repeated in other places this is specifically to find expired tokens or anj registrations.
I have abstracted a bit, should be more clear now, let me know.

packages/server/src/database/migrations/20200509004247_create-notifications.js
```diff
@@ -0,0 +1,17 @@
+export function up(knex) {
+  return knex.schema.createTable('UserNotifications', function (table) {
```
bingen, line 2: If we want to send a notification more than once (like reminders), how would this work? Would we have different `userNotificationTypeId` for each reminder?

packages/server/src/models/objection/User.js
```diff
@@ -50,9 +59,24 @@ export default class User extends BaseModel {
     }
   }
 
+  static async findWithUnverifiedEmail() {
+    const users = await this.query().where({emailVerified: false}).withGraphFetched('[email, emailVerificationToken]')
+    return users.filter(user => !!user.email)
+  }
+
+  async $registeredOnAnj() {
```
bingen, line 67: I’m curious about the naming convention here: does `$` mean that it’s not an static method?

.github/scripts/test/run.sh
```diff
@@ -4,5 +4,7 @@ cp .env.sample $DIR/.env
 cd $DIR
 export SERVER_IMAGE="$1"
 docker-compose up -d
-sleep 10
+sleep 10 # wait for database migrations
```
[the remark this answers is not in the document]
promaty: Good idea, replaced with this guy:
`docker-compose exec court_server /bin/sh -c "npx wait-on http://localhost:\$SERVER_METRICS_PORT"`
(this waits for the server to be actually up not just migrations)

packages/services/src/workers/notification-sender.js
```diff
@@ -0,0 +1,35 @@
...
+  if (!scanner) throw `Notification scanner ${model} not found.`
```
facuspagnuolo, line 21: Instead of throwing we could use `logger` to log an error here cause otherwise, it will stop the execution of all the notification for one wrong DB entry

packages/server/src/models/objection/User.js
facuspagnuolo: In this case I would try to avoid it unless it is extremely necessary

packages/services/src/workers/notification-sender.js
promaty: Good idea, switched to `logger.error` + `return`
Actually I was thinking about services in general, I think it would make more sense to run each service as a separate container. Then I will have more visibility on kubernetes when individual processes are failing and also kubernetes will always try to restart the container if the main process fails. Currently if any worker fails we will not hear about it.

packages/server/src/models/objection/User.js
promaty: okay then, removed it where found one

packages/server/src/models/objection/BaseModel.js
```diff
@@ -18,14 +18,30 @@ export default class BaseModel extends Model {
...
+    return this.query().findById(id)
+  }
+
+  static findOne(args) {
+    return this.query().findOne(args)
+  }
+
   static async findOrInsert(args) {
     let row = await this.query().findOne(args)
     if (!row) row = await this.query().insert(args)
```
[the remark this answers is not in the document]
promaty: haha you really hate that query part :D removed

packages/services/src/models/notification-scanners/NotificationScannerBaseModel.js
```diff
@@ -0,0 +1,34 @@
...
+    return (
+      user?.email &&
+      !user?.notificationSetting?.notificationsDisabled &&
+      (user?.emailVerified || this._sendUnverified)
+    )
+  }
+  get _MINUTES() { return 60 * 1000 }
+  get _HOURS() { return 60 * this._MINUTES }
+  get _DAYS() { return 24 * this._HOURS }
+  get _CLIENT_URL() { return CLIENT_URL }
```
sohkai, line 15: Coming off of the other discussion of defining these as static methods, is there a reason to prefer internal getters over a constant import?

packages/services/src/models/notification-scanners/NotificationScannerBaseModel.js
```diff
@@ -0,0 +1,34 @@
...
+
+  // mandatory methods and functions for extended models
```
sohkai, line 17: In terms of being mandatory, as JS doesn't have a compile-time step for validating abstract classes, you may want to throw an error in the base implementation instead of returning a null default.

packages/services/test/integration/notification-scanners/SubscriptionReminder.js
```diff
@@ -0,0 +1,145 @@
...
+        emailPreferencesUrl: CLIENT_URL
+      },
+      token: user.emailVerificationToken.id
+    })
+    expect(logger.success).to.have.callCount(1)
+  })
+
+  it('should not run scan again before scan period', async () => {
+    await tryRunScanner(logger, notificationTypeModel)
+    expect(logger.success).to.have.callCount(0)
```
sohkai, line 106: This test, and some others relying on the logger being called, are a bit brittle, but not sure if there's something better we can do with a spy instead.

I mention this one especially, because it would pass even if `tryRunScanner()` was completely broken and did nothing.

packages/services/test/integration/notification-scanners/SubscriptionReminder.js
```diff
@@ -0,0 +1,145 @@
...
+  let logger = {}
+  beforeEach(async () => {
+    logger = {
+      success: sinon.fake(),
+      warn: sinon.fake(),
+    }
+  })
+
+  it('should not create a notification for a user with verified address (opted out)', async () => {
+    await userDbCleanup(TEST_ADDR, TEST_EMAIL)
```
sohkai, line 38: It may be useful to create a nested mocha context and use `afterEach()` to do the DB clean up here.

If all the other tests don't rely on carried over state (I think requiring a test with carried over state is usually a red flag in tests though), then maybe we can change the `after()` to be an `afterEach()` :)

packages/services/src/models/notification-scanners/NotificationScannerBaseModel.js
promaty: in every derived class I just do `this` to reach them instead of importing a module every time. Since every derived class needs to define scan periods it felt like this should be part of the base

packages/services/src/models/notification-scanners/NotificationScannerBaseModel.js
promaty: created issue #198

packages/services/test/integration/notification-scanners/SubscriptionReminder.js
promaty: I couldn't think of better spy function.
For that specific test, `tryRunScanner` is supposed to do nothing and return. If there was some bad error then we would see an exception and the whole test would fail

packages/services/test/integration/notification-scanners/SubscriptionReminder.js
promaty: Yes in this case I'm using stateful tests. That's why I'm calling this an integration test because I depend on the database. 
I've added a PR #199 to make this more explicit

PR #157, opened 2020-05-08
Server: handle emails with multiple addresses

promaty:


packages/server/src/validators/UserEmailVerificationTokenValidator.js
```diff
@@ -21,9 +21,6 @@ class UserEmailVerificationTokenValidator extends BaseValidator {
 
   async _validateTokenFormat({ address, token }) {
     const user = await User.query().findOne({address}).withGraphFetched('emailVerificationToken')
-    if (!user) {
-      return this.addError({address: `User ${address} not found`})
```
sohkai, line 25: It's not really wrong, but I'm wondering if we should have used `if else`s instead of returning on every truthy condition?

promaty: yes maybe `if else` would be more clear here, created issue #197

PR #160, opened 2020-05-13
Emails: add `email-verification-reminder` and `missed-reveal` templates

sohkai:
Adds `email-verification-reminder` and `missed-reveal` templates as slight text variations of `email-verification` and `missed-vote`.


emails/templates/email-verification-reminder.js
```diff
@@ -0,0 +1,61 @@
+const { action, base, link, vspace } = require('../template-utils')
```
sohkai, line 1: Please feel free to adjust some of the text if you have any ideas :).

I made this under the assumption that the reminder would contain a new verification JWT, since we have fairly long-lived ones, but wasn't sure after the discussion today.

promaty: Added small text changes to be more distinct form verification email, so users don't get confused in their inbox :)

emails/templates/missed-reveal.js
```diff
@@ -0,0 +1,54 @@
...
+
+      Your vote wasn’t revealed on time. Some of your locked ANJ [1] balance has
+      been forfeit.
+
+      Learn more: {{learnMoreUrl}}
+
+      This service is provided by Aragon One AG [2]. You are receiving this
+      email because you are subscribed to Aragon Court Email Notifications. You
+      can contact us at support@aragon.org if you not longer wish to receive
+      these.
```
promaty, line 42: I guess we could add some helper functions for these parts to avoid repetition in every template

emails/templates/email-verification.js
```diff
@@ -35,15 +35,13 @@ module.exports = function() {
       Verify your email to receive email notifications about important news
       and upcoming tasks.
 
-      To complete the verification process, please click on the button below.
```
sohkai, line 38: Not sure if this change was meant to be made; AFAIK we still have that text in the email?

emails/templates/missed-reveal.js
sohkai: Yes... maybe. The `[2]` part is a bit annoying here to extract this as a helper.

emails/templates/email-verification.js
promaty: Well there is no "button" in the text version of email so I removed it

sohkai: 😅good catch!

PR #161, opened 2020-05-14
Emails: automatic sync

promaty:
This takes care of syncing templates to postmark and assets to storage bucket.
I had to extend the email build command a bit. You can also execute it manually.

### DIFF
--- a/.github/workflows/emails.yml
+++ b/.github/workflows/emails.yml
@@ -1,0 +1,26 @@
+name: Email templates/assets sync
+on:
+  push:
+    branches:
+    - master
+    - development
+    paths:
+    - 'emails/**'
+env:
+  POSTMARK_SERVER_API_TOKEN: ${{secrets.POSTMARK_SERVER_API_TOKEN}}
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+      with:
+        project_id: ${{ secrets.GCP_PROJECT_ID }}
+        service_account_key: ${{ secrets.GCP_SA_KEY }}
+        export_default_credentials: true
+    - run: |
+        cd emails
+        yarn install
+        yarn sync:assets
+        yarn sync:templates
--- a/.github/workflows/emails.yml
+++ b/.github/workflows/emails.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
     - master
-    - development
     paths:
     - 'emails/**'
 env:
--- a/emails/build-emails
+++ b/emails/build-emails
@@ -41,8 +41,7 @@ function renderTemplate(filename) {
 
 async function syncPostmark(files) {
   if (!POSTMARK_SERVER_API_TOKEN) {
-    console.error('You must specify a valid POSTMARK_SERVER_API_TOKEN environment variable.')
-    process.exit(1)
+    throw 'You must specify a valid POSTMARK_SERVER_API_TOKEN environment variable.'
   }
   const client = new postmark.ServerClient(POSTMARK_SERVER_API_TOKEN)
   for (const filename of files) {
@@ -66,9 +65,9 @@ async function syncPostmark(files) {
 function generateTemplates(files) {
   execSync(`rm -rf "${DIST_DIR}"`)
 
+  fs.mkdirSync(DIST_DIR, { recursive: true })
   files.forEach(filename => {
     const { name, template, templateText } = renderTemplate(filename)
-    fs.mkdirSync(DIST_DIR, { recursive: true })
     fs.writeFileSync(path.join(DIST_DIR, name + '.html'), template, 'utf8')
     fs.writeFileSync(path.join(DIST_DIR, name + '.txt'), templateText, 'utf8')
   })
@@ -83,11 +82,8 @@ function printTemplateData(files, filename) {
   const templateFile = files.find(filename => filename === name + '.js')
 
   if (!templateFile) {
-    console.error(
-      `Error: the template ${PRINT_DATA_FOR} couldn’t be found. ` +
+    throw `Error: the template ${PRINT_DATA_FOR} couldn’t be found. ` +
         `Please check the PRINT_DATA_FOR variable.`
-    )
-    process.exit(1)
   }
 
   const createTemplate = require(path.join(
@@ -111,13 +107,13 @@ async function main() {
 
   // Print template data
   if (PRINT_DATA_FOR) {
-    await printTemplateData(files, PRINT_DATA_FOR)
+    printTemplateData(files, PRINT_DATA_FOR)
     return
   }
 
   // sync templates to postmark
   if (SYNC_TEMPLATES) {
-    syncPostmark(files)
+    await syncPostmark(files)
     return
   }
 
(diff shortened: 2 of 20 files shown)


.github/workflows/emails.yml
```diff
@@ -0,0 +1,26 @@
+name: Email templates/assets sync
+on:
+  push:
+    branches:
+    - master
```
sohkai, line 5: The branching is a bit awkward; it would be weird if we had code depending on the templates in `master` but a change in `development` updated them.

emails/templates/missed-vote.js
```diff
@@ -9,6 +9,7 @@ const { accountData } = require('../mock-utils')
 
 module.exports = function() {
   return {
+    subject: 'Your vote wasn’t casted on time',
```
sohkai, line 12:
```suggestion
    subject: 'Your vote wasn’t cast on time',
```

Escaping the wrath of John one `casted` at a time 💨

emails/templates/appeals-opened.js
```diff
@@ -10,6 +10,7 @@ const { accountData } = require('../mock-utils')
 
 module.exports = function() {
   return {
+    subject: 'Appeals are open for Dispute for dispute {{disputeId}}',
```
sohkai, line 13:
```suggestion
    subject: 'Appeals are open for Dispute {{disputeId}}',
```

emails/package.json
```diff
@@ -6,11 +6,13 @@
...
   },
   "scripts": {
     "build": "./build-emails",
     "build:mock": "MOCK_DATA=1 yarn build",
-    "build:postmark": "ASSETS_URL=https://storage.googleapis.com/aragon/court-dashboard/emails/assets yarn build"
-  }
+    "build:postmark": "ASSETS_URL=https://storage.googleapis.com/aragon/court-backend/emails/assets yarn build",
+    "sync:templates": "ASSETS_URL=https://storage.googleapis.com/aragon/court-backend/emails/assets SYNC_TEMPLATES=1 yarn build",
+    "sync:assets": "gsutil cp assets/* gs://aragon/court-backend/emails/assets/"  }
 }
```
sohkai, line 18:
```suggestion
    "sync:assets": "gsutil cp assets/* gs://aragon/court-backend/emails/assets/"
  }
}
```

emails/env.js
```diff
@@ -6,3 +6,9 @@ module.exports.MOCK_DATA =
 
 module.exports.PRINT_DATA_FOR =
   process.env.PRINT_DATA_FOR === undefined ? null : process.env.PRINT_DATA_FOR
+
+module.exports.SYNC_TEMPLATES =
+  process.env.SYNC_TEMPLATES !== undefined && process.env.SYNC_TEMPLATES !== '0'
+
+module.exports.POSTMARK_SERVER_API_TOKEN =
```
sohkai, line 13: Should we document this in the README?

emails/build-emails
```diff
@@ -28,17 +36,41 @@ function renderTemplate(createTemplate, name) {
...
 }
 
 function generateTemplates(files) {
   execSync(`rm -rf "${DIST_DIR}"`)
 
   files.forEach(filename => {
-    const template = require(path.join(__dirname, 'templates', filename))
-    renderTemplate(template, path.basename(filename, '.js'))
+    const { name, template, templateText } = renderTemplate(filename)
+    fs.mkdirSync(DIST_DIR, { recursive: true })
```
sohkai, line 71: Should we lift this to be outside the for loop?

emails/build-emails
```diff
@@ -28,17 +36,41 @@ function renderTemplate(createTemplate, name) {
...
-  fs.mkdirSync(DIST_DIR, { recursive: true })
-  fs.writeFileSync(path.join(DIST_DIR, name + '.html'), template, 'utf8')
-  fs.writeFileSync(path.join(DIST_DIR, name + '.txt'), templateText, 'utf8')
+  return {subject, name, template, templateText}
+}
+
+async function syncPostmark(files) {
+  if (!POSTMARK_SERVER_API_TOKEN) {
+    console.error('You must specify a valid POSTMARK_SERVER_API_TOKEN environment variable.')
+    process.exit(1)
```
sohkai, line 45: Rather than exit here, should we just throw an error?

emails/build-emails
```diff
@@ -5,12 +5,20 @@ const fs = require('fs')
...
+  SYNC_TEMPLATES, 
+  POSTMARK_SERVER_API_TOKEN
+} = require('./env')
 
 const DIST_DIR = path.resolve('dist')
 
-function renderTemplate(createTemplate, name) {
-  let { template, templateText, mockData } = createTemplate()
+function renderTemplate(filename) {
+  const name = path.basename(filename, '.js')
```
sohkai, line 19: It looks like we may want to sanitize the current template names in Postmark first, so we don't get a bunch of duplicates from name mismatches :).

.github/workflows/emails.yml
promaty: Hmm this is a bit tricky, should we have separate templates for master and development then?

emails/build-emails
promaty: I'm following a slightly different alias/naming so I will just manually delete old ones when we merge. You can already see new templates in postmark, I've ran the template sync manually.

emails/build-emails
promaty: Indeed, it was in the loop previously

emails/env.js
promaty: Added!

emails/build-emails
promaty: changed!

.github/workflows/emails.yml
promaty: Changed to only push for master, I think this makes most sense.

sohkai: Yeah, it's definitely a weird place. We can sync to a different postmark server ("staging") for development if we want?

promaty: oh yeah this actually makes most sense, let's create a staging server

emails/build-emails
```diff
@@ -28,17 +36,40 @@ function renderTemplate(createTemplate, name) {
     templateText = handlebars.compile(templateText)(mockData)
   }
 
-  fs.mkdirSync(DIST_DIR, { recursive: true })
-  fs.writeFileSync(path.join(DIST_DIR, name + '.html'), template, 'utf8')
-  fs.writeFileSync(path.join(DIST_DIR, name + '.txt'), templateText, 'utf8')
+  return {subject, name, template, templateText}
```
facuspagnuolo, line 39:
```suggestion
  return { subject, name, template, templateText }
```

emails/build-emails
```diff
@@ -83,7 +111,16 @@ function main() {
...
+  if (SYNC_TEMPLATES) {
+    await syncPostmark(files)
+    return
+  }
+
   // Generate the templates
   generateTemplates(files)
 }
-main()
+main().catch((error) => {
```
facuspagnuolo, line 123:
```suggestion

main().catch((error) => {
```

.github/workflows/emails.yml
promaty: added emails-staging workflow for the postmark server. Also that server will be used by all test isntances on k8s.

PR #165, opened 2020-05-20
Notifications service: first batch of notifications

promaty:
So the flow I used for the test is just basically test the query manually on https://thegraph.com/explorer/subgraph/aragon/aragon-court
and then replace `Network.query()` with the structured result.

You are probably manually testing those queries anyway before writing code so I think this is a good flow to provide some sort of integration testing.


packages/services/src/models/notification-scanners/JurorDrafted.js
```diff
@@ -0,0 +1,38 @@
...
+    const { adjudicationRounds } = await Network.query(query)
+    for (const adjudicationRound of adjudicationRounds) {
+      const { jurors } = adjudicationRound
+      for (const juror of jurors) {
+        notifications.push({ 
+          address: juror.juror.id,
+          details: {
+            emailTemplateModel: {
+              disputeId: juror.id,
+              disputeUrl: `${this._CLIENT_URL}`
```
promaty, line 26: I wasn't sure which URL to use here

packages/services/src/workers/notification-sender.js
```diff
@@ -1,6 +1,7 @@
 import emailClient from '@aragonone/court-backend-shared/helpers/email-client'
 import { UserNotification } from '@aragonone/court-backend-server/build/models/objection'
 import * as notificationScanners from '../models/notification-scanners'
+import { accountData } from '../../../../emails/mock-utils'
```
facuspagnuolo, line 4: should we rename this to `utils` or `helpers`?

packages/services/src/models/notification-scanners/JurorDrafted.js
facuspagnuolo: It should be the dispute URL of the dashboard, e.g. https://court.aragon.org/disputes/13

promaty: added!

packages/services/src/workers/notification-sender.js
promaty: renamed to helpers

packages/services/src/models/notification-scanners/AppealsOpened.js
```diff
@@ -0,0 +1,48 @@
+import NotificationScannerBaseModel from './NotificationScannerBaseModel'
+import Network from '@aragonone/court-backend-server/build/web3/Network'
```
sohkai, line 2: I wonder if we should lift the common dependencies of the services and server out?

In particular, I'm wondering if we should have something like a `db` package that contains all the migrations, configuration, and models. It would be really weird if the server and services each had separate versions of the models.

promaty: Creating issue #196

PR #169, opened 2020-05-25
Notifications service: remaining first batch

promaty:
(DueTasks, AppealsOpened, DisputeRuled, MissedVote, MissedReveal)


packages/services/src/models/notification-scanners/AppealsOpened.js
```diff
@@ -0,0 +1,47 @@
+import NotificationScannerBaseModel from './NotificationScannerBaseModel'
+import Network from '@aragonone/court-backend-server/build/web3/Network'
+
+class AppealsOpened extends NotificationScannerBaseModel {
+  async scan() {
+    let notifications = []
+    const fourDaysBeforeNow = Math.floor((Date.now()-(4*this._DAYS))/1000)
```
promaty, line 7: @bingen you mentioned there should be a better way to get these day periods from blockchain, not sure how

packages/services/src/models/notification-scanners/DisputeRuled.js
```diff
@@ -0,0 +1,45 @@
...
+        jurors
+      } = adjudicationRound
+      for (const juror of jurors) {
+        notifications.push({ 
+          address: juror.juror.id,
+          details: {
+            emailTemplateModel: {
+              disputeId,
+              disputeUrl: `${this._CLIENT_URL}disputes/${disputeId}`,
+              disputeResult: 'Allowed'      // how do I get this result? I only see finalRuling which is an integer
```
promaty, line 33: @facuspagnuolo @bingen , not sure how to get this ruled state as a readable word (it is included in email template)

packages/services/src/models/notification-scanners/AppealsOpened.js
```diff
@@ -0,0 +1,47 @@
...
+    const query = `
+    {
+      adjudicationRounds(where: {stateInt_in: [1,2,3], createdAt_lt: ${fourDaysBeforeNow}}, orderBy: createdAt) {
+        id
+        dispute {
+          id
+        }
+        jurors {
```
promaty, line 15: Should all jurors get AppealsOpened notifications? Should I ignore ones with missed vote/reveal?

packages/services/src/models/notification-scanners/DisputeRuled.js
```diff
@@ -0,0 +1,45 @@
...
+  async scan() {
+    let notifications = []
+    const nineDaysBeforeNow = Math.floor((Date.now()-(9*this._DAYS))/1000)
+    const query = `
+    {
+      adjudicationRounds(where: {state: Ended, createdAt_gt: ${nineDaysBeforeNow}}, orderBy: createdAt) {
+        dispute {
+          id
+        }
+        jurors {
```
promaty, line 14: Should all jurors get DisputeRuled notifications including missed vote/reveal?

packages/services/src/models/notification-scanners/MissedReveal.js
```diff
@@ -0,0 +1,47 @@
...
+        dispute: { id: disputeId },
+        jurors
+      } = adjudicationRound
+      for (const juror of jurors) {
+        notifications.push({ 
+          address: juror.juror.id,
+          details: {
+            emailTemplateModel: {
+              lockedAnjBalanceUrl: 'test',
+              learnMoreUrl: 'test',
```
promaty, line 34: email templates include these locked balance / learn more URLs, not sure where they point, need to confirm with Paty

promaty: edited according to Paty instructions

packages/services/src/models/notification-scanners/AppealsOpened.js
facuspagnuolo: I created this [PR](https://github.com/aragonone/court-backend/pull/170) to provide that information.

You should be able to get a Court wrapper for the network you are working with and call [`config`](https://github.com/aragonone/court-backend/pull/170/files#diff-e097297aec2c8218887bb696a62fbc74R87) passing the [term ID of the dispute](https://github.com/aragon/court-subgraph/blob/master/schema.graphql#L68) we are sending the notification form (we will need to add this join to the query). 

Once you have the config, the values you're looking for are the [round durations](https://github.com/aragonone/court-backend/pull/170/files#diff-e097297aec2c8218887bb696a62fbc74R96). Remember this values are in "court terms", you should multiply them by the [term duration](https://github.com/aragonone/court-backend/pull/170/files#diff-e097297aec2c8218887bb696a62fbc74R75) (measured in seconds not milliseconds)

packages/services/src/models/notification-scanners/AppealsOpened.js
facuspagnuolo: I think it's fine to send a notification to all the drafted jurors even if they fail to commit or reveal their vote

packages/services/src/models/notification-scanners/DisputeRuled.js
facuspagnuolo: Same answer as above

packages/services/src/models/notification-scanners/DueTasks.js
```diff
@@ -0,0 +1,79 @@
...
+     	revealingRounds: adjudicationRounds(where: {stateInt_in: [1,2], createdAt_lt: ${threeDaysBeforeNow}}, orderBy: createdAt) {
+        createdAt
+        dispute {
+          id
+        }
+        jurors (where: {commitment_not: null, revealDate: null}) {
+          juror {id}
+        } 
+      }
+    }
```
facuspagnuolo, line 30: Would it be better to have two scanners for this? One for due commits and another for reveals?

packages/services/src/models/notification-scanners/DisputeRuled.js
facuspagnuolo: We can fetch the [`finalRuling`](https://github.com/aragon/court-subgraph/blob/master/schema.graphql#L70) from the `dispute` object in the query. 

You can follow [this spec](https://github.com/aragon/court-dashboard/blob/94c3d738986903e26b78df9aa8a7f3ff0569c724/src/utils/crvoting-utils.js#L4-L10) to parse the final ruling in the same way we are already doing for the dashboard.

packages/services/src/models/notification-scanners/DueTasks.js
promaty: I was thinking one is better because we can potentially have fewer total emails

facuspagnuolo: Are we mixing due commits and due reveals in the same email for same jurors then?

promaty: Yes, it was like this in the original email template actually:
image

emails/templates/missed-reveal.js
```diff
@@ -21,10 +21,10 @@ module.exports = function() {
           primary: `Your vote wasn’t revealed on time`,
           secondary: `Some of your ${link(
             'locked ANJ balance',
-            '{{lockedAnjBalanceUrl}}'
+            'https://court.aragon.org/dashboard'
```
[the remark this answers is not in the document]
promaty: good point, added as parameter

packages/services/src/models/notification-scanners/DisputeRuled.js
promaty: added

packages/services/src/models/notification-scanners/AppealsOpened.js
promaty: as discussed, added checks with termIds

emails/templates/missed-vote.js
```diff
@@ -48,8 +48,7 @@ module.exports = function() {
     mockData: {
       ...accountData('0xef0f7ecef8385483ac8a2e92d761f571c4b782bd'),
       date: 'Thursday, 17 Dec. 2019',
-      lockedAnjBalanceUrl: '',
-      learnMoreUrl: '',
+      lockedAnjBalanceUrl: 'https://court.aragon.org/dashboard'
```
facuspagnuolo, line 51: We should parameterize this URL as well

packages/services/src/helpers/term-id-getter.js
```diff
@@ -0,0 +1,21 @@
...
+    return currentTerm.sub(commitTerms)
+  }
+  if (state == 'appealing') {
+    return currentTerm.sub(commitTerms).sub(revealTerms) 
+  }
+  if (state == 'commit-reminder') {
+    return currentTerm.sub(commitTerms.add(1).div(2))
+  }
+  if (state == 'reveal-reminder') {
+    return currentTerm.sub(commitTerms).sub(revealTerms.add(1).div(2))
```
facuspagnuolo, line 17: I think you will need to use `add(bn(1))` otherwise `BigNumber` will fail, you can import `bn` from the `shared` subrepo (its a helper to create `BigNumber`s)

packages/services/src/models/notification-scanners/DisputeRuled.js
```diff
@@ -0,0 +1,53 @@
...
+  4: 'InFavor',
+}
+
+class DisputeRuled extends NotificationScannerBaseModel {
+  async scan() {
+    let notifications = []
+    const twoDaysBeforeNow = Math.floor((Date.now()-(2*this._DAYS))/1000)
+    const query = `
+    {
+      disputes(where: {ruledAt_gt: ${twoDaysBeforeNow}}, orderBy: createdAt) {
```
facuspagnuolo, line 17: shouldn't this be `ruledAt_lte`?

emails/templates/missed-vote.js
promaty: It is, this is just used for generating mock templates for preview (you can run `yarn build:mock` in email folder)

packages/services/src/helpers/term-id-getter.js
promaty: Works fine for me, you can import this `term-id-getter.js` file in your local script with ganache node running.

packages/services/src/models/notification-scanners/DisputeRuled.js
promaty: lte will get all disputes in the past, I only want disputes in the last 2 days. (it's the opposite logic createdAt in other scanners)

packages/services/src/helpers/term-id-getter.js
facuspagnuolo: Based on previous experience with BigNumber, I insist that it's safer to use it that way, for consistency and to avoid having issues with different BigNumber versions as this dependency is not usually pinned in the libraries we use

promaty: ok, added!

PR #170, opened 2020-05-26
Shared: Add court wrapper config getter

facuspagnuolo:


packages/shared/models/Court.js
```diff
@@ -80,6 +84,22 @@ module.exports = class {
     return this.instance.getNeededTermTransitions()
   }
 
+  async config(termId = undefined) {
```
promaty, line 87: Maybe rename this to `getTermConfig` or something less ambiguous

promaty:
Can you merge this directly into #169 when it's approved

PR #171, opened 2020-05-26
Notifications service: Notify unverified anj registrations

promaty:
current template render:
Email Subject: **New on Aragon Court: modify your juror notification settings from the Dashboard**
image


promaty:
current template render:
image

PR #172, opened 2020-06-02
Rinkeby: Update contract addresses

facuspagnuolo:


packages/app/src/web3/Network.js
```diff
@@ -4,14 +4,14 @@ import Environment from '@aragonone/court-backend-shared/models/evironments/Brow
...
-  rinkeby: '0x3b86Fd8C30445Ddcbed79CE7eB052fe935D34Fd2',
+  rinkeby: '0x5561f73c3BBe8202F4D7E51aD2A1F22f1E056eFE',
   usability: '0x109dB6047d83f4dd5a8d9da3b9e9228728E3710a',
 }
 
 const ANT = {
   staging: '0x245B220211b7D3C6dCB001Aa2C3bf48ac4CaA03E',
   ropsten: '0x0cb95D9537c8Fb0C947eD48FDafc66A7b72EfC86',
-  rinkeby: '0x5cC7986D7A793b9930BD80067ca54c3E6D2F261B',
+  rinkeby: '0x8cf8196c14A654dc8Aceb3cbb3dDdfd16C2b652D',
```
sohkai, line 14: Just to double check: we don't need to update the ANJ address, right? That's covered by the faucet?

facuspagnuolo: We don't, but actually it's covered by the court instance :)

PR #174, opened 2020-06-03
Chore: Drop usability instance

facuspagnuolo:


packages/app/README.md
```diff
@@ -8,7 +8,6 @@ You can find the following deployed instances
 
 1. [Mainnet](https://court-backend-app.eth.aragon.network/)
 2. [Rinkeby](https://court-backend-app-rinkeby.eth.aragon.network/)
-3. [Usability](https://court-backend-app-usability.eth.aragon.network/)
```
sohkai, line 11: We probably want to update the numbers; a nice trick with Github's markdown is you can use `1.` for all the numbered list items.

facuspagnuolo: Good to know!

PR #176, opened 2020-06-03
Shared: Fix gas price oracle for the testnets

facuspagnuolo:


facuspagnuolo:
haha okey, let's make it 10 gwei

PR #178, opened 2020-06-04
Services: Add auto-reveal double check

facuspagnuolo:


packages/services/src/workers/reveal.js
```diff
@@ -15,8 +15,14 @@ async function reveal(logger, court, reveal) {
...
     await court.revealFor(voteId, juror, outcome, salt)
-    await reveal.$query().update({ revealed: true })
-    logger.success(`Revealed vote #${voteId} for juror ${juror}`)
+    const actualOutcome = await court.getOutcome(voteId, juror)
+
+    if (actualOutcome.toString() === outcome.toString()) {
+      await reveal.$query().update({ revealed: true })
+      logger.success(`Revealed vote #${voteId} for juror ${juror}`)
+    } else {
+      logger.error(`Failed to reveal vote #${voteId} for juror ${juror}. Expected outcome ${outcome}, got ${actualOutcome.toString()}`)
```
promaty, line 24: missed `.toString()` for `outcome`, not sure if important

facuspagnuolo: That one comes from the DB, it's a number... the other one coming from the blockchain is a BN

PR #181, opened 2020-06-04
Emails: minor language fixes

promaty:
Comments from @john-light: 
there are some stray capitalizations "Preliminary Ruling", "Voting" and similarly the button "Appeal Ruling" (where "ruling" should not be title-cased)
we really need to make a writing style guide for Aragon product UIs, so this guidance can be all in one place
but for now I would just suggest the following rules about this: Aragon Court, Dispute #X, and Dashboard or Aragon Court Dashboard are the only words/phrases that need to be capitalized as proper nouns
and the phases ("Appeal phase", "Final ruling" phase, etc) and outcomes ("Refuse to vote", "Block", or "Allow") - but only normal capitalized, not title-cased, and if they are used in the middle of a sentence then put quotes around them to show they are "special" phrases. Bold text can also be used to show the phrase is special e.g. "The final ruling of Dispute #1 is **Refuse to vote**."

### DIFF
--- a/emails/templates/ruled.js
+++ b/emails/templates/ruled.js
@@ -9,7 +9,7 @@ const { accountData } = require('../helpers')
 
 module.exports = function() {
   return {
-    subject: 'Dispute {{disputeId}} has been ruled as “{{disputeResult}}”',
+    subject: 'The final ruling in Dispute #{{disputeId}} is “{{disputeResult}}”',
     template: base(
       {
         title: 'Notifications',
@@ -18,13 +18,14 @@ module.exports = function() {
       `
         ${infobox({
           mode: 'positive',
-          primary: `
-            ${link('Dispute #{{disputeId}}', '{{disputeUrl}}', {
-              nowrap: true,
-            })} has been ruled as “{{disputeResult}}”`,
-          secondary: 'You can see the final result and claim any rewards',
+          primary: `The final ruling in ${link(
+            'Dispute #{{disputeId}}',
+            '{{disputeUrl}}',
+            { nowrap: true }
+          )} has been ruled as “{{disputeResult}}”`,
+          secondary: 'You can now see the final ruling and claim your rewards',
         })}
-        ${action('See result', '{{disputeUrl}}', {
+        ${action('See final ruling', '{{disputeUrl}}', {
           padding: '16px 0 0',
         })}
       `
@@ -34,10 +35,10 @@ module.exports = function() {
 
       Your account {{account}} received a notification on {{date}}:
 
-      Dispute #{{disputeId}} has been ruled as {{disputeResult}}. You can see
-      the final result and claim any rewards.
+      The final ruling in Dispute #{{disputeId}} is “{{disputeResult}}”. 
+      You can now see the final ruling and claim your rewards.
 
-      See result: {{disputeUrl}}
+      See final ruling: {{disputeUrl}}
 
       This service is provided by Aragon One AG [1]. You are receiving this email
       because you are subscribed to Aragon Court Email Notifications. You can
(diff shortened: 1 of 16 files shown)


emails/templates/ruled.js
```diff
@@ -9,7 +9,7 @@ const { accountData } = require('../helpers')
 
 module.exports = function() {
   return {
-    subject: 'Dispute {{disputeId}} has been ruled as “{{disputeResult}}”',
+    subject: 'The final ruling in Dispute #{{disputeId}} is “{{disputeResult}}”',
```
sohkai, line 12:
```suggestion
    subject: 'The final ruling for Dispute #{{disputeId}} is “{{disputeResult}}”',
```

This might sound a bit more natural

emails/templates/ruled.js
```diff
@@ -18,13 +18,14 @@ module.exports = function() {
       `
         ${infobox({
           mode: 'positive',
-          primary: `
-            ${link('Dispute #{{disputeId}}', '{{disputeUrl}}', {
-              nowrap: true,
-            })} has been ruled as “{{disputeResult}}”`,
-          secondary: 'You can see the final result and claim any rewards',
+          primary: `The final ruling in ${link(
```
sohkai, line 21:
```suggestion
          primary: `The final ruling for ${link(
```

emails/templates/ruled.js
```diff
@@ -34,10 +35,10 @@ module.exports = function() {
 
       Your account {{account}} received a notification on {{date}}:
 
-      Dispute #{{disputeId}} has been ruled as {{disputeResult}}. You can see
-      the final result and claim any rewards.
+      The final ruling in Dispute #{{disputeId}} is “{{disputeResult}}”.
```
sohkai, line 38:
```suggestion
      The final ruling for Dispute #{{disputeId}} is “{{disputeResult}}”. 
```

packages/services/src/models/notification-scanners/DisputeRuled.js
```diff
@@ -3,9 +3,9 @@ import Network from '@aragonone/court-backend-server/build/web3/Network'
 const OUTCOMES = {
   0: 'Missing',
   1: 'Leaked',
-  2: 'Refused',
-  3: 'Against',
-  4: 'InFavor',
+  2: 'Refuse to vote',
+  3: 'Block',
+  4: 'Allow',
```
sohkai, line 8: I wonder if these should be in past term, since they're for the ruling?

promaty: I agree, changed!

promaty:
> LGTM.
> 
> Not so sure about the `textFooter()` change as we lose the ability to render more links, but I think it's OK as we're not serving that many text emails.

We can just add more below, starting with[2]

packages/services/src/models/notification-scanners/DisputeRuled.js
john-light: I disagree. The ruling is not "Allowed", it is "Allow" (and same for the other rulings). I see nothing wrong with a statement such as:

> The final ruling in Dispute #0 is: Allow

or if you are writing/speaking in the past tense:

> After two rounds of deliberation, in their final ruling jurors a final ruling to **Allow** the disputed action.

promaty: Hmm but it's also past in the dashboard itself so if we change we should do it everywhere at the same time.
image

promaty: image

john-light: I did not notice that before but if that's the case then yes I think we should be consistent about what the ruling outcomes are and keep the rulings in the present tense ("Allow", "Block", "Refuse to vote")

PR #185, opened 2020-06-05
Services: updated anj notification script

promaty:


packages/server/scripts/notify-unverified-anj-registrations.js
```diff
@@ -15,17 +15,26 @@ dotenv.config()
...
     const { email: { email }, address } = user
-    await emailClient.sendEmailWithTemplate({
-      To: email,
-      TemplateAlias: 'notification-settings-announcement',
-      TemplateModel: {
-        dashboardUrl: 'https://court.aragon.org/dashboard',
-      },
-    })
+    if (emails[email]) continue
+    emails[email] = 1
```
sohkai, line 22: For what it's worth... this is where a `Set()` might be better :). We may also want to lowercase if we haven't already.

promaty: Fixed!

PR #187, opened 2020-06-09
Server: Fix reveal show endpoint

facuspagnuolo:


packages/server/README.md
```diff
@@ -106,6 +106,8 @@ All the provided endpoints are `Content-Type: application/json`
           "voteId": "0", 
           "disputeId": "0", 
           "roundNumber": "0",
+          "revealed": true,
+          "failedAttempts": 2,
```
sohkai, line 110: Not sure if this needs to be returned.

Do we also need the `revealed` information returned, since wouldn't this be detectable on-chain?

fabriziovigevani: Yes, probably is not needed for the dashboard at least, since we would be doing this request only on the committing phase. We would just be checking if there's a record or not for the given juror and voteId

facuspagnuolo: Cool! Let's remove it then

facuspagnuolo: Fixed now! LMKYT

PR #190, opened 2020-06-17
Services: notification sender fix

promaty:
For the past 2 days notification sender stopped working because postmark was returning an error for one of the emails. I guess the user verified the email and then deleted afterwards and postmark marked it as non-working.

image

Just before merging to master I should delete all unsent notifications otherwise users will receive a wave of notifications at the same time.

### DIFF
--- a/packages/server/src/models/objection/UserNotification.js
+++ b/packages/server/src/models/objection/UserNotification.js
@@ -1,6 +1,5 @@
 import BaseModel from './BaseModel'
-const MINUTES = 60 * 1000
-const HOURS = 60 * MINUTES
+import { HOURS } from '@aragonone/court-backend-shared/helpers/times'
 
 export default class UserNotification extends BaseModel {
   static get tableName() {
--- a/packages/server/src/models/objection/UserNotification.js
+++ b/packages/server/src/models/objection/UserNotification.js
@@ -1,4 +1,6 @@
 import BaseModel from './BaseModel'
+const MINUTES = 60 * 1000
+const HOURS = 60 * MINUTES
 
 export default class UserNotification extends BaseModel {
   static get tableName() {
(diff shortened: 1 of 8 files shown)


sohkai:
> Just before merging to master I should delete all unsent notifications otherwise users will receive a wave of notifications at the same time.

Trying to understand what this means. It's because a redeployment will restart the timer on all the notification checks (even ones that were previously unsent)?

promaty:
> > Just before merging to master I should delete all unsent notifications otherwise users will receive a wave of notifications at the same time.
> 
> Trying to understand what this means. It's because a redeployment will restart the timer on all the notification checks (even ones that were previously unsent)?

Basically there are a bunch of notifications in the past several days which are marked as `sentAt=null`, that's the condition I check  when trying to send stuff. Now that I think of it I should probably also do a check `createdAt > past day` or something like that

promaty:
Added a condition for only sending notifications created in the last hour.

packages/server/src/models/objection/UserNotification.js
```diff
@@ -1,4 +1,6 @@
 import BaseModel from './BaseModel'
+const MINUTES = 60 * 1000
+const HOURS = 60 * MINUTES
```
facuspagnuolo, line 3: We can probably move this to a times helper module, I think we have declared these in multiple places already

packages/server/src/models/objection/UserNotification.js
```diff
@@ -25,4 +26,10 @@ export default class UserNotification extends BaseModel {
       },
     }
   }
+
+  static findUnsent() {
+    return this.query()
+      .whereNull('sentAt')
+      .andWhere('createdAt', '>', new Date(Date.now()-HOURS))
```
sohkai, line 33: Haha, we can do this... but let's use something like [dayjs](https://github.com/iamkun/dayjs) (or [date-fns](https://date-fns.org/), but we primarily use dayjs on the frontend because it's much, much lighter).

It's not a huge problem in the backend because we don't have to be concerned about different operating environments (e.g. client timezones), but usually still better to use a tried and true library for these kinds of manipulations.

Issue #138, opened 2020-04-07
PR backlog review comments / questions


sohkai:
Left comments on a number of older PRs:

- [x] [Server: Improve create reveal endpoint](https://github.com/aragonone/court-backend/pull/126#pullrequestreview-389051582) (question)
- [x] [Server: Implement basic metrics](https://github.com/aragonone/court-backend/pull/120#pullrequestreview-389067059) (questions and suggestions)
- [ ] [Service: Improve monitor keeper](https://github.com/aragonone/court-backend/pull/113#pullrequestreview-389100452) (suggestions)
- [x] [Shared: Add monitoring setup](https://github.com/aragonone/court-backend/pull/111#pullrequestreview-389128124) (typo nitpick)
- [ ] [Chore: Improve dockerization](https://github.com/aragonone/court-backend/pull/89#pullrequestreview-389130000) (suggestions)
- [ ] `packages/shared/models/evironments` directory is typoed

PR #193, opened 2020-06-25
Service: Improve monitor keeper

facuspagnuolo:
Fixes https://github.com/aragonone/court-backend/issues/138

### DIFF
--- a/packages/services/src/models/Etherscan.js
+++ b/packages/services/src/models/Etherscan.js
@@ -34,8 +34,8 @@ export default class Etherscan {
   }
 
   _getBaseUrl(network) {
-    if (network === 'rinkeby' || network === 'ropsten') return BASE_URL.replace('<network>', network)
-    else if (network === 'staging') return BASE_URL.replace('<network>', 'rinkeby')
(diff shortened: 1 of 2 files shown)


packages/services/src/models/Etherscan.js
```diff
@@ -34,8 +34,8 @@ export default class Etherscan {
   }
 
   _getBaseUrl(network) {
-    if (network === 'rinkeby' || network === 'ropsten') return BASE_URL.replace('<network>', network)
-    else if (network === 'staging') return BASE_URL.replace('<network>', 'rinkeby')
+    if (network === 'rinkeby' || network === 'ropsten') return BASE_URL.replace('<network>', `api-${network}`)
+    else if (network === 'staging') return BASE_URL.replace('<network>', 'api-rinkeby')
```
sohkai, line 38: So I assume this wasn't working at all for the test networks—should we have seen error logs?

facuspagnuolo: Both work actually, but the one specified in the documentation is the one with the `api-` prefix :)

PR #198, opened 2020-06-26
Notification service: throwing errors on unimplemented base class methods

promaty:


packages/services/src/models/notification-scanners/NotificationScannerBaseModel.js
```diff
@@ -16,19 +16,29 @@ export default class NotificationScannerBaseModel {
...
+   * return [
+   *   { 
+   *     address: null,
+   *     details: {
+   *       emailTemplateModel: {}
+   *     }
+   *   }
+   * ]
+   */
+  async scan() { throw 'subclass responsibility' }
```
sohkai, line 30: This works, but it's generally better to wrap it in a `new Error()` so the stack trace and etc. get included.

facuspagnuolo: Yup, let's always use `throw Error` which I think it's the shortcut for `throw new Error()`

facuspagnuolo: Yup, let's always use `throw Error` which I think it's the shortcut for `throw new Error()`

promaty: Added!

PR #205, opened 2020-09-02
notifications: changed to hash client url in emails

promaty:


sohkai (approved):
🙏 let's get this deployed!

PR #207, opened 2020-09-03
Reveals: Handle reveal period

facuspagnuolo:


packages/services/src/workers/reveal.js
```diff
@@ -16,6 +20,16 @@ async function reveal(logger, court, reveal) {
   const { voteId, juror, outcome, salt, failedAttempts } = reveal
   try {
     logger.info(`Revealing vote #${voteId} for juror ${juror}`)
+    const { disputeId, roundNumber } = reveal
+    const { canReveal, expired } = await court.getRevealStatus(disputeId, roundNumber)
+
+    if (expired) {
+      await reveal.$query().update({ expired: true })
+      return logger.error(`Reveal vote #${voteId} for juror ${juror} is expired!`)
```
sohkai, line 28:
```suggestion
      return logger.error(`Reveal vote #${voteId} for juror ${juror} has expired!`)
```

packages/services/src/workers/reveal.js
```diff
@@ -16,6 +20,16 @@ async function reveal(logger, court, reveal) {
...
+    } else if (!canReveal) {
+      return logger.warn(`Cannot reveal round #${roundNumber} for dispute #${disputeId} yet`)
```
sohkai, line 30: Just wondering, would this be somewhat noisy? This would continuously log from the time we got the commit to the time we reveal, right?

sohkai (approved):
LGTM, let's get it deployed!

packages/services/src/workers/reveal.js
facuspagnuolo: Yes, let's see how it goes, and adjust it later

PR #214, opened 2020-09-07
chore: fixed lerna versioning, migrated to yarn

promaty:
Just adding some improvements from ipfs-pinner.

You can now use `yarn` for everything, it will install single `yarn.lock` at the top with all the packages so we don't have any versioning errors anymore.

For publishing new version just run:
`yarn lerna version patch`, it will create a tag, update package.json files, push to git in one command.


promaty:
@facuspagnuolo seems that test is failing because I removed the `web3-eth-abi`.
When I didn't remove it `yarn install` throws this error:
`error web3-eth-abi@1.0.0: The engine "node" is incompatible with this module. Expected version ">=8.0.0 <=11.15.0". Got "12.18.2"`
It's a bit weird because you should see same error on `npm install`, but it ignores it?

Can you take a look? I tried upgrading to `"web3-eth-abi": "^1.2.11"` locally, but the test is still failing for me

lerna.json
```diff
@@ -2,5 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "independent"
+  "version": "0.2.18",
+  "npmClient": "yarn",
+  "useWorkspaces": true
```
facuspagnuolo, line 7: can't we use yarn workspaces directly instead of lerna?

Dockerfile
```diff
@@ -24,9 +24,10 @@ RUN mkdir -p /packages/shared
 COPY ./packages/shared/package.json /court-backend/packages/shared/package.json
 
 # install dependencies
-RUN npm install
-RUN npx lerna bootstrap
+COPY ./yarn.lock /court-backend/yarn.lock
+RUN yarn install
+RUN yarn lerna link
```
facuspagnuolo, line 29: running simply `yarn` should be enough I think

README.md
```diff
@@ -67,8 +67,8 @@ First, open a separate terminal, clone this repo and install dependencies:
 ```bash
   git clone https://github.com/aragonone/court-backend/
   cd court-backend
-  npm i
-  npx lerna bootstrap
+  yarn install
+  yarn lerna link
```
facuspagnuolo, line 71: same above

facuspagnuolo (commented):
@promaty I don't think the error is related to `web3-eth-abi`.

It fails cause it is not finding an artifacts file which makes me think we should probably define all the artifacts-related dependencies for nohoist:
- @aragon/minime
- @aragon/court
- @aragonone/precedence-campaign-arbitrable

lerna.json
promaty: yarn is for installing packages, lerna is for bumping versions

Dockerfile
promaty: yeah it's fine, but I prefer install less ambiguous

promaty:
Hm it seems to work fine if I don't user workspaces. I would still prefer to use it though, then we can have single lock file, looking into it

promaty:
@facuspagnuolo , seems that adding nohoist for @aragon/minime is enough, thanks!

packages/shared/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "@aragonone/court-backend-shared",
   "version": "0.2.18",
   "private": true,
+  "workspaces": {
+    "nohoist": ["@aragon/court"]
```
facuspagnuolo, line 6: we should add the other ones I pointed out, clearly we will need them even tho those are not being used in the tests

lerna.json
facuspagnuolo: Can't we use [`yarn version`](https://classic.yarnpkg.com/en/docs/cli/version/)? 
I don't think it is necessary to have lerna installed for it tbh

packages/shared/package.json
promaty: added

lerna.json
promaty: Cannot figure this out for the moment, yarn alone doesn't work for me. Let's create an issue for this

Dockerfile
sohkai: Ah, I think Facu meant that `lerna link` is unnecessary since we're using yarn workspaces now.

promaty: Yes but I wasn't able to figure out right now why yarn couldn't link and version packages by itself, so opened another issue for this

PR #216, opened 2020-09-08
services: added contract monitor

promaty:


facuspagnuolo (approved):
LGTM!

packages/services/src/models/Etherscan.js
```diff
@@ -16,7 +17,7 @@ export default class Etherscan {
     return bn(response.result)
   }
 
-  async getTransactionsFrom(address, fromBlock) {
+  async getTransactionsFrom(address, fromBlock = 0) {
     const path = `${this.url}module=account&action=txlist&address=${address}&startblock=${fromBlock}&sort=desc`
```
sohkai, line 21: In terms of this failing, I saw that we throw an error. Are these pushed up as generic errors that we can then alert on?

I'm mostly worried about the case where etherscan is either unresponsive or blocks our API key, and we don't notice.

promaty: It will bubble up all the way to worker and if worker is not executing for a while I already have an alert for this in k8s manifests `CourtBackendWorkerNotRunning`

sohkai: Gotcha, in terms of code style, I wonder if it's worth explicitly catching and throwing the error in the worker itself so it's more clear how we'd be affected by this type of error.

Or perhaps this is a futile effort and we should just be cognizant that internal calls may bubble errors and cause the worker to die (kind of like Java and uncaught exceptions)?

promaty: Well we have this catch block here at the top, which updates a metric and prints the exact error:
https://github.com/aragonone/court-backend/blob/development/packages/services/bin/worker.js#L35
So if etherescan is not responding we will have a worker error metric and see the exact error in the logs

sohkai: Ok, all good! Didn't realize we had those there :).

PR #218, opened 2020-09-10
App: Add manual email client

promaty:
My first attempt at React :D


packages/app/src/actions/emails.js
```diff
@@ -0,0 +1,48 @@
...
+      try {
+        const decoder = new TextDecoder('utf-8')
+        const response = await fetch(`${SERVER_URL}/emails`, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json'
+          },
+          credentials: 'include',
+          body: JSON.stringify(params),
+        })
```
facuspagnuolo, line 19: What do you think if we use [this](https://github.com/aragonone/court-backend/blob/development/packages/app/src/web3/Server.js) guy?

packages/app/src/actions/emails.js
```diff
@@ -0,0 +1,48 @@
...
+        const reader = response.body.getReader()
+        for (;;) {
+          const { done, value } = await reader.read()
+          dispatch(EmailActions.receveEmailLog(decoder.decode(value)))
+          if (done) {
+            dispatch(EmailActions.testEmailSent(true))
+            break
+          }
+        }
```
facuspagnuolo, line 28: What about

```js
const reader = response.body.getReader()

for (let done, value; !done;) {
 ({ done, value } = await reader.read())
  dispatch(EmailActions.receveEmailLog(decoder.decode(value)))
}

dispatch(EmailActions.testEmailSent(true))
```

packages/app/src/actions/emails.js
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }
+  },
+
+  resetEmailLogs(){
+    return { type: ActionTypes.RESET_EMAIL_LOGS }
+  },
+
+  receveEmailLog(emailLog) {
+    return { type: ActionTypes.RECEIVE_EMAIL_LOG, emailLog }
```
facuspagnuolo, line 40: if it's an email-related action already, we can probably rename `emailLog` to `log`

packages/app/src/reducers/emails.js
```diff
@@ -0,0 +1,19 @@
...
+const initialState = { emailLogs: [] }
+
+const EmailReducer = (state = initialState, action) => {
+  switch(action.type) {
+    case ActionTypes.RECEIVE_EMAIL_LOG:
+      const emailLogs = [...state.emailLogs, action.emailLog]
+      return Object.assign({}, state, { emailLogs })
+    case ActionTypes.RESET_EMAIL_LOGS:
+      return Object.assign({}, state, { emailLogs: [] })
+    case ActionTypes.RECEIVE_TEST_EMAIL_STATUS:
```
facuspagnuolo, line 12: In the context of the store this should be related to the action, maybe simply `enable` instead of `testEmailSent`?

packages/server/src/controllers/EmailsController.js
```diff
@@ -0,0 +1,46 @@
...
+export default {
+  async send(req, res) {
+    const { TemplateModel, testEmail, testEmailSent } = req.body
+    let users = []
+    if (testEmailSent) {
+      users = await User.findWithoutDisabledNotifications()
+    }
+    else if (testEmail) {
+      users = [{
+        email: { email: testEmail },
```
facuspagnuolo, line 14: What do you think about having simply a recipient list? if there is any we send to those, otherwise we send to jurors

packages/server/src/controllers/EmailsController.js
```diff
@@ -0,0 +1,46 @@
...
+        address: 'TESTADDRESS'
```
facuspagnuolo, line 15: `0x0000000000000000000000000000000000000000` maybe?

packages/app/src/components/core/Navbar.react.js
```diff
@@ -26,6 +26,7 @@ export default class Navbar extends React.Component {
           <Link to="/subscribers">Subscribers</Link>
           <Link to="/periods">Periods</Link>
           <Link to="/anj-balances">ANJ</Link>
+          { !admin.id && <Link to="/admin">Login</Link> }
```
facuspagnuolo, line 29: Let's remove this line, it is not necessary to show people this is available

packages/app/src/components/admin/EmailsForm.react.js
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  onStateChange() {
+    const { admin: { email }, emails: { emailLogs, testEmailSent } } = Store.getState()
+    if (!this.state.testEmail && email) this.setState({ testEmail: email })
+    this.setState({ emailLogs, testEmailSent })
+  }
+
+  componentWillUnmount() {
+    this.unsubscribe()
```
facuspagnuolo, line 123: are you sure redux doesn't do this automatically?

facuspagnuolo (commented):
Looking good! 💪

packages/app/src/actions/emails.js
promaty: I tried, but axios don't do streaming in the frontend

packages/app/src/actions/emails.js
promaty: nice changed.

packages/app/src/actions/emails.js
promaty: good point, renamed

packages/app/src/components/admin/EmailsForm.react.js
promaty: I get console errors after changing the page if I don't do this:
`Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in EmailsForm (created by Context.Consumer)`

packages/app/src/components/core/Navbar.react.js
promaty: ok removed

packages/server/src/controllers/EmailsController.js
promaty: Yeah makes more sense, added

packages/server/src/controllers/EmailsController.js
promaty: I did it first, but then I thought maybe sending to all jurors by default is not a good idea, we should be more explicit about it.
Changed variable name to `notifyAllJurors`

packages/app/src/reducers/emails.js
promaty: Replaced with
```js
    case ActionTypes.RECEIVE_EMAIL_FORM_ENABLED:
      return Object.assign({}, state, { enabled: action.enabled })
```

facuspagnuolo (approved):

PR #221, opened 2020-10-21
Shared: added typescript

promaty:
Added typescript for shared packages.

Only converted the times module for now, we can keep adding others one at a time.


promaty:
Moving this to https://github.com/aragon/protocol-backend

